PR #73, opened 2015-02-22
Another code cleanup

fcanc:
Since my last cleanup can't be merged, I did most of the job again. 
I reused autoPEP8 since last time something went lost during merge.
In addition I reviewed some texts and I also fixed a bug that I caught during some testing (the Auto-Tag feature froze in GUI mode).
Some considerations about the ordering of imports. I know the rule (first the standards, then 3rd party and your stuff last) is not strict and under some circumstances it could also break something, but I took the time to test a lot and everything seems fine. Also having commented some unused imports (and they were a lot!) works well for me. I hope I don't break anything on the other OS as well.


davide-romanini:
Now it should be ok. :-)

fcanc:
Thanks! I'm wondering what is the best practice for pushing changes in the comicapi package, since it has its own repository, should I change that instead? And will it update here (and in comicstreamer as well)?

davide-romanini:
I've used the subtree split strategy for the comicapi folder (https://makingsoftware.wordpress.com/2013/02/16/using-git-subtrees-for-repository-separation/).
The main reason for the split is providing a mechanism to share that part of code with comicstreamer, that previously was just copied over and gone out of sync. Actually, as far as I know the "real work" should happen here, or in the comicstream repo and then pushed on the comicapi repo.
It's the first time I use git subtree so I'm not sure if it's the right solution, but at the moment seems enough to make some progress :-)

fcanc:
But what if different changes happen on both of the repositories? It'll be great if everything will synch automatically, but as of now it could bring a lot of confusion (just my 2 cents).

PR #74, opened 2015-03-21
Bugs in move2folder.py script

Alkpone:
See commit details to see the errors

### DIFF
--- a/scripts/move2folder.py
+++ b/scripts/move2folder.py
@@ -77,7 +77,7 @@ def main():
     comic_list = []
     max_name_len = 2
     for filename in filelist:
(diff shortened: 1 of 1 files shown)


fcanc:
Thanks for the bug-fix! I noticed also the other scripts are broken, hope this will apply to those as well. Unfortunately I can't take a look right now as I'm in the middle of a format/reinstall...

davide-romanini:
I suspect it could be better tackled in the ComicArchive implementation, after all the nocover is an optional argument (`default_image_path=None`), we probably should check for None.

Alkpone:
Indeed, it is a quick fix, I guess there could be some refactoring, but it doesn't yet add value.

Issue #65, opened 2015-02-03
Remove import of WebPImagePlugin


davide-romanini:
_From [g...@geekraver.com](https://code.google.com/u/110414153140219914094/) on January 22, 2015 16:20:23_

What version of ComicTagger are you using? Top of tree as of 22 Jan 2015. On what operating system (Mac, Linux, Windows)? What version? Mac GUI or command line? GUI What steps will reproduce the problem? Running the installed package doesn't have this issue, but if I run from the command line, ComicTagger thinks I have no PIL installed because of a failed import of WebPImagePlugin. This import is used in three places to tell if PIL is available, but the import itself is not used. It is enough to just use "rom PIL import Image" to test for PIL.

_Original issue: http://code.google.com/p/comictagger/issues/detail?id=65_

PR #78, opened 2015-06-23
Remove WebP imports from PIL initialization

ghost:
Remove extraneous WebP imports that can break the app on hosts with no WebP support.


davide-romanini:
Should resolve #65

lordwelch:
fixed in #214

Issue #60, opened 2015-02-03
Images-pages not loaded from archive


davide-romanini:
_From [fabiocan...@gmail.com](https://code.google.com/u/100979784519175915875/) on December 14, 2014 14:40:35_

What version of ComicTagger are you using? 1.1.15-beta On what operating system (Mac, Linux, Windows)? What version? Mac OS X Yosemite 10.10.1 GUI or command line? GUI What steps will reproduce the problem? 1. Load a comic archive whose image filenames contain the character ' (apostrophe)
2. Look at the image preview and the page count on the left, and the Pages tab as well What is the expected output? What do you see instead? Said images aren't loaded, resulting in a comic archive with none or just a few pages and compromising the auto-tag and auto-identify search functions Please provide any additional information below. I'm not sure but I suspect this behaviour applies to other characters too

_Original issue: http://code.google.com/p/comictagger/issues/detail?id=60_

lordwelch:
Current alpha release does not have this issue.

PR #79, opened 2015-06-24
Fix issue #60: Fallback to PNG instead of TIFF. 

ghost:
TIFF conversion doesn't work well with some Python installations like Conda on Mac.


davide-romanini:
Applied same fix on latest develop. Closing.

PR #83, opened 2016-07-28
Fixing missing default image bug

goodevilgenius:
I had a problem when using the API from an external script, which was throwing this error:

```
  File "/usr/lib/python2.7/site-packages/comictagger-1.1.15b0-py2.7.egg/comicapi/comicarchive.py", line 648, in __init__
    with open(fname, 'rb') as fd:
TypeError: coercing to Unicode: need string or buffer, NoneType found
````

This fixes that error


goodevilgenius:
Accidentally added other commits

Issue #88, opened 2017-04-03
RAR library support


lacqui:
Comictagger currently relies on [unrar](https://pypi.python.org/pypi/unrar/) which seems to be lacking in support.

davide-romanini:
Closing because for now I prefer keeping libunrar.

PR #89, opened 2017-04-03
Use recently-maintained 'rarfile' instead of 'unrar'.

lacqui:
Fix issue #88 by changing to [rarfile](https://pypi.python.org/pypi/rarfile/).  Additionally remove source for unused UnRAR2


davide-romanini:
Nice work but:
 - rarfile implements rar structure (in pure python)
 - rarfile implements rar file extraction of uncompressed files (in pure python)
 - rarfile still relies on rar binary to extract compressed files

So I'm still in doubt if rarfile is really a better choice. Currently I've spent a lot of time to better integrate libunrar in the build system (on a feature branch). I know we still require rar binary for actually writing tags, that I admit is the core business of comictagger.
For now I would prefer sticking with unrar / libunrar.
In a future version maybe we'll rework on it.

PR #125, opened 2019-03-17
This will allow for the rename-function to also move files, in Linux

ScuttleSE:
In the rename-template, use an absolute path to move files

for example

`/Comics/%publisher%/%series%/%series% %issue%`


lordwelch:
This feature has been implemented in #129

PR #129, opened 2019-09-03
Renaming

lordwelch:
Moves to Python format strings for renaming, handles directory
structures, moving of files to a destination directory, sanitizes
file paths with pathvalidate and takes a different approach to
smart filename cleanup using the Python string.Formatter class

Moving to Python format strings means we can point to python
documentation for syntax and all we have to do is document the
properties and types that are attached to the GenericMetadata class.

Switching to pathvalidate allows comictagger to more simply handle both
directories and symbols in filenames.

The only changes to the string.Formatter class is:
1. format_field returns
an empty string if the value is none or an empty string regardless of
the format specifier.
2. _vformat drops the previous literal text if the field value
is an empty string and lstrips the following literal text of closing
special characters.

Includes #188


davide-romanini:
@lordwelch, the feature looks nice, I wanted something like that for a long time, thank you for your work. :-)
I noticed that entering an invalid python format string, the application exits without further notice. It would be nice to catch the error and notifying the user so it can fix it.
Can you introduce such a change in the PR yourself?
Thank you.

lordwelch:
@davide-romanini, thank you for the response.
I completely forgot about accounting for invalid format strings. I guess I should actually learn how to write tests for my code :-)
I should have another commit in sometime tomorrow

lordwelch:
Template error checking has been implemented.

lordwelch:
Rebased off of latest changes

PR #131, opened 2019-09-03
Allow extended selection in the page list editor

lordwelch:


lordwelch:
Rebased off of latest changes

PR #133, opened 2019-09-04
Add publisher and imprint handling

lordwelch:
Imprint handling has been added to utils and uses a subclassed dict to return tuples for imprint matching may not be the best idea but it works for now.

Add settings option auto_imprint
Add cli flag -a, --auto-import


lordwelch:
Rebased off of latest changes

PR #134, opened 2019-09-10
Move to python requests module

lordwelch:
requests is much simpler and fixes all ssl errors.
Comic Vine now requires a unique useragent string


lordwelch:
Rebased off of latest changes

PR #138, opened 2019-09-22
Make pip release, tweak search request

beville:
Sorry, two things at once, I'm not great with github.

1.  Main thing is fixing the ability to create pip-able zip files.  By running "make pydist", a zip file will be created in "piprelease" subfolder.  This can be used with pip3.  i.e.:
`pip3 install comictagger-1.1.32rc1.zip`
 It can also be uploaded to pypi.org
Note:  The format of the version string had to be changed to remove the '-' to conform to PEP440.

2.  Changed the Comic Vine search request URL to ask for a maximum of 100 items at once, instead of the default of 10.  This should result in less Comic Vine hits.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,6 @@
 include README.md
 include release_notes.txt
-include requirements.txt
+include setup_requirements.txt
 include unrar/*
 recursive-include scripts *.py *.txt
 recursive-include desktop-integration *
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ pydist:
 	make clean
 	mkdir -p piprelease
 	rm -f comictagger-$(VERSION_STR).zip
-	python setup.py sdist --formats=zip  #,gztar
+	python3 setup.py sdist --formats=zip  #,gztar
 	mv dist/comictagger-$(VERSION_STR).zip piprelease
 	rm -rf comictagger.egg-info dist
 		
--- a/PKG-INFO
+++ b/PKG-INFO
@@ -1,0 +1,38 @@
+Metadata-Version: 1.1
+Name: comictagger
+Version: 1.1.32rc1
+Summary: A cross-platform GUI/CLI app for writing metadata to comic archives
+Home-page: https://github.com/davide-romanini/comictagger
+Author: ComicTagger team
+Author-email: comictagger@gmail.com
+License: Apache License 2.0
+Description: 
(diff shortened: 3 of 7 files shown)


beville:
A few more notes:

Earlier this year I got this set up, experimented with getting a release uploaded to pypi.org, and then I had a HDD crash before I could commit anything.  The months slipped away, and I never got back to it.

So, with the recent activity, I got it working again, and managed to commit changes this time :-)

Also, I just took the liberty of uploading 1.1.32rc1 to pypi.org.  

The process is very simple:

Prereq: Install twine:
`pip3 install twine`

Build dist package:
`make pydist`

Upload new release:
`twine upload  piprelease/*`

(You need a login with admin permissions for comictagger project, of course.  Davide, if you don't have the comictagger project credentials still, I can send them to you.  Or if you create a new personal account, I can add you as an admin.)

Once up, the whole world can just: 
`pip3 install comictagger`

On my (now-defucnt) Mac I was just using this version, with python3, pyqt5, etc provided by homebrew (https://brew.sh/).  On Ubuntu (and other linux systems), it just seems to work. Haven't tested on Windows lately, but I I remember once I had Python3 installed, it wasn't much work to run from this package as well.

davide-romanini:
Nice to have, and furthermore it should be easy to automatically upload it during release thanks to [travis PyPI deployment](https://docs.travis-ci.com/user/deployment/pypi/).
I've some doubt related to the correct handling of libunrar though, because it's binary and depends on the specific platform. As far as I know in this case a wheel should be more appropriate, but need some more investigation on the topic.
By the way, I'll look into this.

beville:
For binary distributions for Mac and Windows a wheel would probably be good, but beyond my capacity at the moment, as I'm pretty much using Linux boxes almost exclusively lately.  I think there should always be a source-only distro too, though. 

The source package setup.py  will attempt to compile libunrar during the pip install.  This works on Ubuntu (and I presume any Linux distro).  If the compile fails, the install just proceeds without it. Even without the auto-built libunrar, the GUI (on any platform) will allow the user to specify the location of the library file (along with hints on where to download).

The main thing that I punted on with the pip install is the desktop integration to put shortcut/icon files in the right place.

All this automated build stuff is very cool, I'm learning a lot watching you do this.

davide-romanini:
I splitted the work in #168 and #164. I made a testing [1.3.0a0 release](https://pypi.org/project/comictagger/1.3.0a0/).
Closing this PR.

PR #195, opened 2021-09-20
Make archiver plugable

lordwelch:
This is a relatively simple plugin system for archivers built using [entry points].

Adding support for plugins loaded from a folder can be implemented later. See [flake8] for inspiration.

[entry points]: https://packaging.python.org/en/latest/specifications/entry-points/
[flake8]: https://flake8.pycqa.org/en/latest/user/configuration.html#using-local-plugins


lordwelch:
Should also define plugin folders for each OS

Issue #143, opened 2019-09-27
Comicvine Search Results issues


jdinda70:
Running 1.2.0+2 Develop Branch on Windows

2 issues:

1.

When I search for: "Spider-Man" in the Series it returns the 1990 volume of Spider-Man from Comicvine, but not the 2016 or 2019 volumes of Spider-Man (of course along with a bunch of other volumes with Spider-Man in the name).

They are returned via the comic vine website when I search and restrict the search to "Volumes".

Searching for "Spider" or "Man" separately also does not return the 2016 or 2019 volumes.

2.

When I search for "X-Men" in the series it does not return ANY with just the volume name "X-Men" from ComicVine (there are 24 volumes total, 4 from Marvel: 1991, 2004, 2010, & 2013)

They are returned via the comic vine website when I search and restrict the search to "Volumes".

Searching for "X" or "Men" separately also does not return these volumes.

There maybe other similar issues, but these are two i have seen recently.

davide-romanini:
Seems related to #112. 
Currently I'm trying to stabilize and automate the build and release process. After that I hope to have a better base to test and fix all these issues.

beville:
Just looked into it, and this is the result of aggressive pruning of search results.  Last year's change to Comic Vine's search required some changes to keep  the amount of results from being too overwhelming.  The fact that the results are sorted according some undefined 'relevance' algorithm is helpful. 

The fetching is stopped when any of a few criteria are matched.  One is if the number of words in the result volume name is greater than the original request, plus a fixed threshold, which is hard-coded at 3 (named as "result_word_count_max"). This idea was a shot in the dark, but has seemed to work pretty well so far.

The code in question is in "searchForSeries()" in comicvinetalker.py

I found two simple (but maybe not perfect) ways to handle this:

1) Use the patch in #138 to allow more results per fetch (from 10 (default) to 100).

 2) Increase or allow the user to configure the threshold that is used to calculate result_word_count_max

And less simple: 

3) Rework the result-limiting algorithm in searchForSeries().  I think that CV Scraper plugin for ComicRack did something quite different, but I can't remember what.

Regardless, I still recommend upping the search result limit since it reduces the number of hits to Comic Vine's API.

jdinda70:
Guys ... thanks for looking into this.  Don't stress.  It would be great to find a solution at some point but it sounds like it's a challenging issue.  Thanks for pointing out the file/function where this code is. I don't know anything about Python, but I'll see if I can figure it out.

[...]

PR #200, opened 2021-09-28
Series sorting filtering

thFrgttn:
Because additional series results are now returned due to #143 the series selection window can with a large number of results that are not usually sorted in a useful way.

I've created 3 settings that can help finding the corect series quickly
1. use the publisher black list - can be toggled from the series selction screen, as well as a setting for is default behaviour
2. a setting mak the result initally sortted by start year insead of the default no of issues
3. a setting to initially put exact and near matches at the top of the list

### DIFF
--- a/comictaggerlib/ui/settingswindow.ui
+++ b/comictaggerlib/ui/settingswindow.ui
@@ -263,6 +263,40 @@
                 </property>
                </widget>
               </item>
+             <item>
+              <widget class="Line" name="line_4">
+               <property name="sizePolicy">
+                <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+                 <horstretch>0</horstretch>
+                 <verstretch>0</verstretch>
+                </sizepolicy>
+               </property>
+               <property name="orientation">
+                <enum>Qt::Horizontal</enum>
+              </property>
+            </widget>
+            </item>
+              <item>
+               <widget class="QCheckBox" name="cbxUseBlackFilter">
+                <property name="text">
+                 <string>Always use Publisher Blacklist on 'manual' searches</string>
+                </property>
+               </widget>
+              </item>
+              <item>
+               <widget class="QCheckBox" name="cbxSortByYear">
+                <property name="text">
+                 <string>Initally sort Series search results by Starting Year instead of No. Issues</string>
+                </property>
+               </widget>
+              </item>
+              <item>
+               <widget class="QCheckBox" name="cbxExactMatches">
+                <property name="text">
+                 <string>Initally show Series Name exact matches first</string>
+                </property>
+               </widget>
+              </item>
              </layout>
             </item>
            </layout>
--- a/comictaggerlib/ui/settingswindow.ui
+++ b/comictaggerlib/ui/settingswindow.ui
@@ -286,7 +286,7 @@
               <item>
                <widget class="QCheckBox" name="cbxSortByYear">
                 <property name="text">
-                 <string>Initally sort Series search results by Starting Year</string>
+                 <string>Initally sort Series search results by Starting Year instead of No. Issues</string>
                 </property>
                </widget>
               </item>
--- a/comictaggerlib/ui/settingswindow.ui
+++ b/comictaggerlib/ui/settingswindow.ui
@@ -263,6 +263,40 @@
                 </property>
                </widget>
               </item>
+             <item>
+              <widget class="Line" name="line_4">
+               <property name="sizePolicy">
+                <sizepolicy hsizetype="Expanding" vsizetype="Fixed">
+                 <horstretch>0</horstretch>
+                 <verstretch>0</verstretch>
+                </sizepolicy>
+               </property>
+               <property name="orientation">
+                <enum>Qt::Horizontal</enum>
+              </property>
+            </widget>
+            </item>
+              <item>
+               <widget class="QCheckBox" name="cbxUseBlackFilter">
+                <property name="text">
+                 <string>Always use Publisher Blacklist on 'manual' searches</string>
+                </property>
+               </widget>
+              </item>
+              <item>
+               <widget class="QCheckBox" name="cbxSortByYear">
+                <property name="text">
+                 <string>Initally sort Series search results by Starting Year</string>
+                </property>
+               </widget>
+              </item>
+              <item>
+               <widget class="QCheckBox" name="cbxExactMatches">
+                <property name="text">
+                 <string>Initally show Series Name exact matches first</string>
+                </property>
+               </widget>
+              </item>
              </layout>
             </item>
            </layout>
--- a/comictaggerlib/ui/volumeselectionwindow.ui
+++ b/comictaggerlib/ui/volumeselectionwindow.ui
@@ -148,6 +148,13 @@
            </property>
           </widget>
          </item>
+        <item>
+         <widget class="QCheckBox" name="cbxFilter">
+          <property name="text">
+           <string>Apply Publishers Filter</string>
+          </property>
+         </widget>
+        </item>
          <item>
           <widget class="QDialogButtonBox" name="buttonBox">
            <property name="orientation">
--- a/comictaggerlib/ui/volumeselectionwindow.ui
+++ b/comictaggerlib/ui/volumeselectionwindow.ui
@@ -151,7 +151,10 @@
         <item>
          <widget class="QCheckBox" name="cbxFilter">
           <property name="text">
-           <string>Apply Publishers Filter</string>
+           <string>Filter Publishers</string>
+          </property>
+          <property name="toolTip">
+           <string>Filter the publishers based on the publisher blacklist.</string>
           </property>
          </widget>
         </item>
--- a/comictaggerlib/volumeselectionwindow.py
+++ b/comictaggerlib/volumeselectionwindow.py
@@ -336,12 +336,12 @@ class VolumeSelectionWindow(QtWidgets.QDialog):
             return
 
         self.cv_search_results = self.search_thread.cv_search_results
-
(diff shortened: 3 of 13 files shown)


comictaggerlib/ui/settingswindow.ui
```diff
@@ -263,6 +263,40 @@
...
+              </property>
+            </widget>
+            </item>
+              <item>
+               <widget class="QCheckBox" name="cbxUseBlackFilter">
+                <property name="text">
+                 <string>Always use Publisher Blacklist on 'manual' searches</string>
+                </property>
+               </widget>
+              </item>
```
lordwelch, line 285: This should be next to the blacklist entry, it makes more sense to have all of the blacklist settings together.
It may make sense to rewrite the Identifier tab to be 'Search' instead.

comictaggerlib/ui/volumeselectionwindow.ui
```diff
@@ -148,6 +148,13 @@
            </property>
           </widget>
          </item>
+        <item>
+         <widget class="QCheckBox" name="cbxFilter">
+          <property name="text">
+           <string>Apply Publishers Filter</string>
```
lordwelch: We should be consistent on naming I would probably go with filter

comictaggerlib/volumeselectionwindow.py
```diff
@@ -32,9 +32,11 @@
 from .matchselectionwindow import MatchSelectionWindow
 from .coverimagewidget import CoverImageWidget
 from comictaggerlib.ui.qtutils import reduceWidgetFontSize, centerWindowOnParent
+
+from comictaggerlib import settings
```
lordwelch, line 36: This import isn't used

comictaggerlib/volumeselectionwindow.py
```diff
@@ -124,13 +126,18 @@ def __init__(self, parent, series_name, issue_number, year, issue_count,
         self.cover_index_list = cover_index_list
         self.cv_search_results = None
 
+        self.use_blackList = self.settings.use_publisher_blacklist_for_manual
```
lordwelch: Calling this always use blacklist/filter is more descriptive and makes it so that you don't need the context of the blacklist help text to understand the usage of the setting

comictaggerlib/volumeselectionwindow.py
```diff
@@ -334,6 +345,40 @@ def searchComplete(self):
             return
 
         self.cv_search_results = self.search_thread.cv_search_results
+        # filter the blacklisted publishers if setting set
+        if self.use_blackList:
+            try:
+                publisher_blacklist = set([s.strip().lower() for s in self.settings.id_publisher_blacklist.split(',')])
```
lordwelch: So I just discovered that there are [set comprehensions](https://dmerej.info/blog/post/hello-flake8/) you just use curly braces instead of straight braces.
```suggestion
                publisher_blacklist = {s.strip().lower() for s in self.settings.id_publisher_blacklist.split(',')}
```

comictaggerlib/volumeselectionwindow.py
```diff
@@ -334,6 +345,40 @@ def searchComplete(self):
...
+                # use '' as publisher name if None
+                self.cv_search_results = list(filter(lambda d: ('' if d['publisher'] is None else str(d['publisher']['name']).lower()) not in publisher_blacklist, self.cv_search_results))
```
lordwelch, line 353: There's nothing wrong with this but I count 8 different things going on in this statement and maintaining a loop constructing a new list is much simpler

comictaggerlib/volumeselectionwindow.py
```diff
@@ -334,6 +345,40 @@ def searchComplete(self):
...
+        if self.settings.sort_series_by_year:
+            try:
+                self.cv_search_results = sorted(self.cv_search_results, key = lambda i: (str(i['start_year']), str(i['count_of_issues'])), reverse=True)
+            except:
+                print('bad data error sorting results by start_year,count_of_issues')
+        else:
+            try:
+                self.cv_search_results = sorted(self.cv_search_results, key = lambda i: str(i['count_of_issues']), reverse=True)
+            except:
+                print('bad data error sorting results by count_of_issues')
```
lordwelch, line 370: natsort handles different types well also operator.itemgetter makes simple keys much easier

```suggestion
        # pre sort the data, so that we can put exact matches first
        # sort by start_year if set
        if self.settings.sort_series_by_year:
            try:
                self.cv_search_results = natsorted(self.cv_search_results, key=itemgetter('start_year', 'count_of_issues'), reverse=True)
            except:
                print('error sorting results by start_year,count_of_issues')
        else:
            try:
                self.cv_search_results = natsorted(self.cv_search_results, key=itemgetter('count_of_issues'), reverse=True)
            except:
                print('error sorting results by count_of_issues')
```

comictaggerlib/volumeselectionwindow.py
```diff
@@ -375,9 +420,9 @@ def searchComplete(self):
 
             row += 1
 
-        self.twList.resizeColumnsToContents()
+        # redundant
+        #self.twList.resizeColumnsToContents()
```
lordwelch:
```suggestion
```

lordwelch (changes requested):
Hello, I put a few notes in on what I noticed, also could you squash your commits it makes for a much cleaner git history.

comictaggerlib/volumeselectionwindow.py
thFrgttn: Your correct, settings are imported 4 lines above

comictaggerlib/volumeselectionwindow.py
thFrgttn: This now fails if the start_year has a non numerical value -
Comic VInes has a 'Captain Marvel' series with a start year of '1976?'

comictaggerlib/volumeselectionwindow.py
thFrgttn: This fails sorting records containing None values
when you search Comic Vine for a series called 'Excalibur', one of the results is a bed result that just has None in the other fields

comictaggerlib/volumeselectionwindow.py
thFrgttn: this altered to deal with bad data, which is why it got so busy
my only question about using a loop, is which is faster

comictaggerlib/volumeselectionwindow.py
lordwelch: What causes the failure this is creating a set for the publisher filter, where does the year come into play?

comictaggerlib/volumeselectionwindow.py
lordwelch: This is true I only tested when comparing strings and None which are comparable however numeric types are not comparable with None

comictaggerlib/ui/volumeselectionwindow.ui
thFrgttn: Changed it to 'Filter Publishers' and added a tooltip mentioning the publisher blackllist

comictaggerlib/volumeselectionwindow.py
thFrgttn: agreed

comictaggerlib/volumeselectionwindow.py
lordwelch: Technically it is faster to use a comprehension ([blog](https://towardsdatascience.com/speeding-up-python-code-fast-filtering-and-slow-loops-8e11a09a9c2f) note that all of the code snippets are juypeter notebooks and it is not mentioned that they are not pure python), however the current cap is hardcoded at 500 and even if you remove the cap "batman" only has 1842 series the filter takes 0.0013023210003666463 seconds, the entire searchComplete function takes 1.603541956999834 seconds

If you are going to talk about speed this is not the function to do it at.

When opening a cached search for the first time versus clicking on the filter checkbox to re-run just the sorting and filtering something makes it go faster so the total for the searchComplete function is 0.113114669999959 seconds without the filtering which is still 100 times longer than the filtering takes on 1842 series.

I'll merge it either way but next time someone takes a look at filtering a loop is much easier to parse through in my experience

lordwelch:
Also binaries built using pyinstaller are uploaded for macos and windows if you look under the checks

comictaggerlib/volumeselectionwindow.py
thFrgttn: Think I reported this wrong, I must have tested bad code
(need to stop dong things late a night)

comictaggerlib/ui/settingswindow.ui
thFrgttn: I don't have an editor for qt ui files installed, and the layout of Identifier tab is just a bit too complex to edit from editting the xml text, and not something I'm going to be able to do

lordwelch: The editor is qt designer I'll work something in tonight.
Here is a windows build I've used in the past as otherwise you need to install the full Qt buildsystem
https://build-system.fman.io/qt-designer-download

lordwelch: Here are two patches one is just the ui and consistency the other renames blacklist to filter
[0001-Update-the-blacklist-settings-no-rename.patch.txt](https://github.com/comictagger/comictagger/files/7264268/0001-Update-the-blacklist-settings-no-rename.patch.txt)
[0001-Update-the-blacklist-settings-rename-to-filter.patch.txt](https://github.com/comictagger/comictagger/files/7264269/0001-Update-the-blacklist-settings-rename-to-filter.patch.txt)

lordwelch:
@thFrgttn If you squash your commits I will merge this pull request

PR #216, opened 2022-03-02
remove macos version from artifact name

nicerloop:
Just keep the platform architecture, not the build host version, in the produced artifact.
It makes the artifact name more stable and in line with windows platform.
It also simplifies automatic update and artifact fetching for homebrew https://brew.sh


lordwelch:
I am hesitant to merge this as the build host version is valuable information as in general anything built on a newer version of OSX will not work on an older version.

Also is there a reason why brew is not building from source?
rar has always been an optional feature as you have to pay for a license to modify archives, the only other binary dependencies are pyqt and pillow which are already in brew, at which point you would just run `pip install .` to install ComicTagger.

lordwelch:
Closing, feel free to comment if you want to discuss further

PR #225, opened 2022-03-18
Added 7zip support thru py7zr.

MichaelFitzurka:
Tweaked save of archive file and images in comicarchive.

### DIFF
--- a/comicapi/comicarchive.py
+++ b/comicapi/comicarchive.py
@@ -14,6 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import py7zr
 import zipfile
 import os
 import struct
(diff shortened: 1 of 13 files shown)


scripts/name_fixer.py
```diff
@@ -54,7 +54,9 @@ def calculate_rename(ca, md, settings):
 
     new_ext = None  # default
     if settings.rename_extension_based_on_archive:
-        if ca.isZip():
+        if ca.isSevenZip():
+            new_ext = ".cb7"
+        elif ca.isZip():
```
lordwelch: Let's not edit the scripts unless we are also going to update them to python3, it makes it simpler to trace their history when they aren't muddled by ineffective edits.

scripts/remove_ads.py
```diff
@@ -51,7 +51,7 @@ def main():
     for filename in filelist:
 
         ca = ComicArchive(filename, settings.rar_exe_path)
-        if (ca.isZip or ca.isRar()) and ca.hasMetadata(style):
+        if (ca.isSevenZip() or ca.isZip() or ca.isRar()) and ca.hasMetadata(style):
```
lordwelch: Same as above

comicapi/comicarchive.py
```diff
@@ -1037,7 +1153,10 @@ def applyArchiveInfoToMetadata(self, md, calc_page_sizes=False):
                         data = self.getPage(idx)
                         if data is not None:
                             try:
-                                im = Image.open(io.StringIO(data))
+                                if isinstance(data, bytes):
+                                    im = Image.open(io.BytesIO(data))
+                                else:
+                                    im = Image.open(io.StringIO(data))
```
lordwelch: What does this have to do with 7z?

comicapi/comicarchive.py
```diff
@@ -50,6 +51,109 @@ class MetaDataStyle:
...
+    def setArchiveComment(self, comment):
+        return False
+
+    def readArchiveFile(self, archive_file):
+        data = ""
+        try:
+            with py7zr.SevenZipFile(self.path, 'r') as zf:
+                data = zf.read(archive_file)[archive_file].read()
+        except py7zr.Bad7zFile as e:
+            print("bad zipfile [{0}]: {1} :: {2}".format(e, self.path,
```
lordwelch: This should reference 7zip and not zipfile

comicapi/comicarchive.py
```diff
@@ -50,6 +51,109 @@ class MetaDataStyle:
...
+                archive_file), file=sys.stderr)
+            raise IOError
+        except Exception as e:
+            print("bad zipfile [{0}]: {1} :: {2}".format(e, self.path,
```
lordwelch: same as above

lordwelch (changes requested):

comicapi/comicarchive.py
MichaelFitzurka: copy past error - I'll fix

comicapi/comicarchive.py
MichaelFitzurka: I can't remember why I made this change - i think it had to do with  a png file, but am not 100%.
I am 100% sure it wasn't 7zip - so I'll remove

comicapi/comicarchive.py
MichaelFitzurka: CPE

lordwelch (approved):

PR #230, opened 2022-03-25
Revise file sort in archive

rhaussmann:
I'm on macOS and the IC (ignore case) flag doesn't appear to work. The last file ends up first in order and the first file ends up (usually) second to last. I combined the original lower() function with the natsort. I also tried updating natsort >=4.0, which should allow locale() or humansort() to work on macOS, but I had no luck. However, sort() works just on the files I tested.

### DIFF
--- a/comictaggerlib/comicvinetalker.py
+++ b/comictaggerlib/comicvinetalker.py
@@ -740,14 +740,15 @@ class ComicVineTalker(QObject):
         for d in div_list:
             if 'class' in d.attrs:
                 c = d['class']
-                if ('imgboxart' in c and
-                        'issue-cover' in c and
-                        d.img['src'].startswith("http")
-                   ):
-
-                    covers_found += 1
-                    if covers_found != 1:
+                if 'imgboxart' in c and 'issue-cover' in c:
+                    if d.img['src'].startswith("http"):
+                        covers_found += 1
+                        if covers_found != 1:
                             alt_cover_url_list.append(d.img['src'])
+                    elif d.img['data-src'].startswith("http"):
+                        covers_found += 1
+                        if covers_found != 1:
+                            alt_cover_url_list.append(d.img['data-src'])
 
         return alt_cover_url_list
 


lordwelch:
Can you provide a file that reproduces this issue? You can download the CBZ at [this](https://launchpad.net/acbf) site and just rename the pages. I'm just trying to find out what caused this as I can't find any filenames that produce the issue your having.

Here is a trivial script which lists all the pages in a comic archive. It needs either comictagger installed or to be placed next to comictagger.py
[list.txt](https://github.com/comictagger/comictagger/files/8310605/list.txt)
I've tested this both on the latest commit with every version of natsort between 5.1.0 and 8.1.0 and they all report the same output and the macOS binary for 1.3.0 and all report the correct sort order for ascii text.

Also this pull request cannot be merged as the `signed` keyword argument was removed in version 6.0.0 https://github.com/SethMMorton/natsort#dropped-deprecated-apis

Can you report your environment?
locale:
```sh
$ defaults read -g AppleLocale # if on macOS
$ locale # if on linux
```
or in python

```python
>>> import locale
>>> locale.getlocale()
```
OS and version info:
```sh
$ cat /etc/os-release # if on linux
$ sw_vers # if on macOS
$ python -VV
$ pip list
```

rhaussmann:
I got the current version working when **natsort** is set to 8.1.0 but the minimum version in the **requirements.txt** isn't specified. It seems like it would be better to specify in the requirements, unless that screws something else up. 
Here's my specs FYI.
>>> import locale
>>> locale.getlocale()
('en_US', 'UTF-8')

(venv) ➜  comictagger git:(natsort_fix) ✗ sw_vers
ProductName:    macOS
ProductVersion: 12.2.1
BuildVersion:   21D62

(venv) ➜  comictagger git:(natsort_fix) ✗ python -VV
Python 3.9.10 (main, Jan 15 2022, 11:48:04) 
[Clang 13.0.0 (clang-1300.0.29.3)]

(venv) ➜  comictagger git:(natsort_fix) ✗ pip list
Package                   Version
------------------------- ---------
altgraph                  0.17.2
beautifulsoup4            4.10.0
certifi                   2021.10.8
cffi                      1.15.0
charset-normalizer        2.0.7
configparser              5.0.2
idna                      3.3
macholib                  1.15.2
natsort                   3.5.2
Pillow                    8.4.0
pip                       21.3.1
pycparser                 2.20
pyinstaller               4.3
pyinstaller-hooks-contrib 2021.3
PyPDF2                    1.24
PyQt5                     5.15.5
PyQt5-Qt5                 5.15.2
PyQt5-sip                 12.9.0
PyQt5-stubs               5.15.2.0
requests                  2.26.0
setuptools                57.0.0
soupsieve                 2.2.1
unrar                     0.3
unrar-cffi                0.2.2
urllib3                   1.26.7
wheel                     0.36.2

Thanks for your work on this BTW. Since the slow demise of ComicRack, this project has been a great organizational resource.

lordwelch:
Could you squash your commits and then force push this branch? It will make this change easier to follow once merged.
Examples can be found here https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git

Issue #232, opened 2022-03-26
Advertisment vs Advertisement


MichaelFitzurka:
The schema: https://github.com/anansi-project/comicinfo/blob/main/schema/v2.0/ComicInfo.xsd
Has the page type as: "Advertisment" (no e)
Our code has: "Advertisement" (with e)

I noticed the issue when loading an archive with the schema's value (no e), and the application crashes on save (as below).

It is a simple change and I can create a pull request - but then all old archives with the old spelling could similarly crash.

A cleaner approach might be to default the old value to the new one silently, but it would require an archive save to fix permanently in the archive's xml file.

Thoughts?

Traceback (most recent call last):
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 2063, in fileListSelectionChanged
    self.loadArchive(fi.ca)
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 2081, in loadArchive
    self.actualLoadCurrentArchive()
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 610, in actualLoadCurrentArchive
    self.pageListEditor.setData(self.comic_archive, self.metadata.pages)
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/pagelisteditor.py", line 275, in setData
    item = QListWidgetItem(self.listEntryText(p))
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/pagelisteditor.py", line 287, in listEntryText
    text += " (" + self.pageTypeNames[page_dict['Type']] + ")"
KeyError: 'Advertisment'
Traceback (most recent call last):
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 2063, in fileListSelectionChanged
    self.loadArchive(fi.ca)
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 2067, in loadArchive
    self.clearForm()
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/taggerwindow.py", line 748, in clearForm
    self.pageListEditor.setData(self.comic_archive, self.metadata.pages)
  File "/home/michael.fitzurka/Projects/comictagger/comictaggerlib/pagelisteditor.py", line 271, in setData
    self.listWidget.itemSelectionChanged.disconnect(self.changePage)
[...]

PR #238, opened 2022-03-31
Displaying invalid value with Error indicator

MichaelFitzurka:
That way the user can see what is the invalid value and has the option to leave it or change it.


MichaelFitzurka:
Fixes #232

PR #240, opened 2022-03-31
Adding web link convenience button to open a valid url value in a browser window.

### LINKED ISSUE #239: Feature request: Button to open up web link url in a browser
I'll add a pull request shortly.

MichaelFitzurka:
resolves #239

### DIFF
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -25,12 +25,13 @@ import json
 import webbrowser
 import re
 import pickle
-import requests
 #import signal
 
 from PyQt5 import QtCore, QtGui, QtWidgets, uic
 from PyQt5 import QtNetwork
 from PyQt5.QtCore import QUrl
+
+from urllib.parse import urlparse
 
 #from comicarchive import ComicArchive
 #from pageloader import PageLoader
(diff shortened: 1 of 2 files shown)


comictaggerlib/taggerwindow.py
```diff
@@ -1368,6 +1373,19 @@ def removeCredit(self):
             self.twCredits.removeRow(row)
         self.setDirtyFlag()
 
+    def openWebLink(self):
+        if self.leWebLink is not None:
+            web_link = "{0}".format(self.leWebLink.text()).strip();
```
lordwelch: this can be shortened to `self.leWebLink.text().strip()`

comictaggerlib/taggerwindow.py
```diff
@@ -1368,6 +1373,19 @@ def removeCredit(self):
...
+            if len(web_link) > 0:
```
lordwelch: You can also use `if web_link:` it's mostly equivalent and accounts for None values but shouldn't matter in this case

comictaggerlib/taggerwindow.py
```diff
@@ -1368,6 +1373,19 @@ def removeCredit(self):
...
+                try:
+                    response = requests.get(web_link)
```
lordwelch: This response isn't used.

comictaggerlib/taggerwindow.py
```diff
@@ -25,6 +25,7 @@
 import webbrowser
 import re
 import pickle
+import requests
```
lordwelch: requests isn't actually used in this change

lordwelch (changes requested):

comictaggerlib/taggerwindow.py
MichaelFitzurka: done

comictaggerlib/taggerwindow.py
MichaelFitzurka: done - removed variable.

comictaggerlib/taggerwindow.py
MichaelFitzurka: It is used to validate the url.

comictaggerlib/taggerwindow.py
MichaelFitzurka: Done.

comictaggerlib/taggerwindow.py
lordwelch: urllib is in the standard library and has a parse method to validate URLs the code currently used actually makes a network request and then throws away the response which is not desirable

MichaelFitzurka: urlparse passes mangled urls like: **https://https://https://www.foo.bar** as valid.
But is good enough.
Changes made.

lordwelch (approved):

Issue #243, opened 2022-04-03
AttributeError: 'SevenZipArchiver' object has no attribute 'get_filename_list'


MichaelFitzurka:
All method names changed, but not for 7zip - incoming pull request.

lordwelch:
I new I forgot something! Thanks for catching it

PR #244, opened 2022-04-03
method renamed to match new changes.

MichaelFitzurka:
Fix for 7zip.
Caught an issue while saving with black_and_white field as well.


MichaelFitzurka:
fixes #243

PR #277, opened 2022-04-22
Adding CommunityRating.

### LINKED ISSUE #258: Anansi Project: Schema v2.0: CommunityRating
Likewise, we are almost fully compliant with the Anansi v2.0 schema.
The only outlier here is the CommunityRating value.

``` xml
<xs:element minOccurs="0" maxOccurs="1" name="CommunityRating" type="Rating" />

    <xs:simpleType name="Rating">
        <xs:restriction base="xs:decimal">
            <xs:minInclusive value="0"/>
            <xs:maxInclusive value="5"/>
            <xs:fractionDigits value="2"/>
        </xs:restriction>
    </xs:simpleType>
```

Or, 0 to 5 with 2 decimal points.
The decimal points gets changed to only 1 in the v2.1 schema.

This would be easy enought to add to the Notes tab, after the User Rating.
The value is available on some ComicVine pages:
ex. https://comicvine.gamespot.com/wonder-woman-89-home/4000-55260/

``` html
<span class="[average-score]()">4.1 stars</span>
```

MichaelFitzurka:
Fixes #258

### DIFF
--- a/comictaggerlib/settingswindow.py
+++ b/comictaggerlib/settingswindow.py
@@ -105,7 +105,7 @@ Accepts the following variables:
 {black_and_white} (boolean)
 {page_count}     (integer)
 {maturity_rating}    (string)
-{community_rating}     (float)
+{community_rating}     (string)
 {story_arc}      (string)
 {series_group}   (string)
 {scan_info}      (string)
--- a/comictaggerlib/settingswindow.py
+++ b/comictaggerlib/settingswindow.py
@@ -105,6 +105,7 @@ Accepts the following variables:
 {black_and_white} (boolean)
 {page_count}     (integer)
 {maturity_rating}    (string)
+{community_rating}     (float)
 {story_arc}      (string)
 {series_group}   (string)
 {scan_info}      (string)
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -888,7 +888,7 @@ Please choose options below, and select OK.
         md.notes = self.teNotes.toPlainText()
         md.maturity_rating = self.cbMaturityRating.currentText()
 
-        md.community_rating = utils.xlate("{:.1f}".format(self.dsbCommunityRating.value()))
+        md.community_rating = utils.xlate(self.dsbCommunityRating.cleanText())
         if md.community_rating == "0.0":
             md.community_rating = None
 
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -768,6 +768,11 @@ Please choose options below, and select OK.
         assign_text(self.teTeams, md.teams)
         assign_text(self.teLocations, md.locations)
 
+        try:
+            self.dsbCommunityRating.setValue(float(md.community_rating))
+        except:
+            self.dsbCommunityRating.setValue(0.0)
+
         if md.format is not None and md.format != "":
             i = self.cbFormat.findText(md.format)
             if i == -1:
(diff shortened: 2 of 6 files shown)


comictaggerlib/settingswindow.py
```diff
@@ -105,6 +105,7 @@
 {black_and_white} (boolean)
 {page_count}     (integer)
 {maturity_rating}    (string)
+{community_rating}     (float)
```
lordwelch: This states float however the community_rating attribute is never a float, it only is ever a string or None

comictaggerlib/taggerwindow.py
```diff
@@ -883,6 +888,10 @@ def form_to_metadata(self):
         md.notes = self.teNotes.toPlainText()
         md.maturity_rating = self.cbMaturityRating.currentText()
 
+        md.community_rating = utils.xlate("{:.1f}".format(self.dsbCommunityRating.value()))
```
lordwelch: I recommend using the [cleanText](https://doc.qt.io/qt-5/qdoublespinbox.html#cleanText-prop) function as it should handle all text validation

comictaggerlib/ui/TemplateHelp.ui
```diff
@@ -91,6 +91,7 @@ tr:nth-child(even) {
      &lt;tr&gt;&lt;td&gt;{black_and_white}&lt;/td&gt;&lt;td&gt;boolean&lt;/td&gt;&lt;/tr&gt;
      &lt;tr&gt;&lt;td&gt;{page_count}&lt;/td&gt;&lt;td&gt;integer&lt;/td&gt;&lt;/tr&gt;
      &lt;tr&gt;&lt;td&gt;{maturity_rating}&lt;/td&gt;&lt;td&gt;string&lt;/td&gt;&lt;/tr&gt;
+     &lt;tr&gt;&lt;td&gt;{community_rating}&lt;/td&gt;&lt;td&gt;float&lt;/td&gt;&lt;/tr&gt;
```
lordwelch: Same as above

comictaggerlib/ui/taggerwindow.ui
```diff
@@ -981,6 +981,44 @@
...
+                   <double>0.000000000000000</double>
+                  </property>
+                  <property name="maximum">
+                   <double>5.000000000000000</double>
+                  </property>
+                  <property name="singleStep">
+                   <double>0.100000000000000</double>
+                  </property>
+                 </widget>
+                </item>
```
lordwelch: I would recommend unsetting sizePolicy and acceptDrops as they don't provide any use over the default.
I would also recommend changing maximumSize to minimumSize with a width of 80 so that it will match the criticalRating box above it.

```diff
diff --git a/comictaggerlib/ui/taggerwindow.ui b/comictaggerlib/ui/taggerwindow.ui
index 0d67f8c..90d5c31 100644
--- a/comictaggerlib/ui/taggerwindow.ui
+++ b/comictaggerlib/ui/taggerwindow.ui
@@ -990,21 +990,12 @@
                 </item>
                 <item row="4" column="1">
                  <widget class="QDoubleSpinBox" name="dsbCommunityRating">
-                  <property name="sizePolicy">
-                   <sizepolicy hsizetype="Fixed" vsizetype="Fixed">
-                    <horstretch>0</horstretch>
-                    <verstretch>0</verstretch>
-                   </sizepolicy>
-                  </property>
-                  <property name="maximumSize">
+                  <property name="minimumSize">
                    <size>
                     <width>80</width>
-                    <height>16777215</height>
+                    <height>0</height>
                    </size>
                   </property>
-                  <property name="acceptDrops">
-                   <bool>false</bool>
-                  </property>
                   <property name="decimals">
                    <number>1</number>
                   </property>
```

lordwelch (changes requested):
I would probably store the CommunityRating as a float instead of a string as a string invites invalid data to be stored/processed

comictaggerlib/ui/taggerwindow.ui
MichaelFitzurka: That's the XML that Qt5 Designer generated - but I will change.

lordwelch: Hmm... must be a difference between the version of Qt Designer that I have, but it indicated that the properties were set manually I just unset them in Qt Designer

lordwelch (approved):

PR #323, opened 2022-06-29
Generalise cache db

mizaki:
This is the start. More of a test PR and so you can spot anything early. I may have added source in places it's not really needed?

I left the CV<var> names alone for now. Probably not worth the hassle to change them?

### DIFF
--- a/comictaggerlib/comiccacher.py
+++ b/comictaggerlib/comiccacher.py
@@ -28,10 +28,10 @@ from comictaggerlib.settings import ComicTaggerSettings
 logger = logging.getLogger(__name__)
 
 
-class ComicVineCacher:
+class ComicCacher:
     def __init__(self) -> None:
         self.settings_folder = ComicTaggerSettings.get_settings_folder()
-        self.db_file = os.path.join(self.settings_folder, "cv_cache.db")
+        self.db_file = os.path.join(self.settings_folder, "comic_cache.db")
         self.version_file = os.path.join(self.settings_folder, "cache_version.txt")
 
         # verify that cache is from same version as this one
@@ -72,42 +72,45 @@ class ComicVineCacher:
         # create tables
         with con:
             cur = con.cursor()
-            # name,id,start_year,publisher,image,description,count_of_issues
+            # source_name,name,id,start_year,publisher,image,description,count_of_issues
             cur.execute(
                 "CREATE TABLE VolumeSearchCache("
                 + "search_term TEXT,"
-                + "id INT,"
+                + "id INT NOT NULL,"
                 + "name TEXT,"
                 + "start_year INT,"
                 + "publisher TEXT,"
                 + "count_of_issues INT,"
                 + "image_url TEXT,"
                 + "description TEXT,"
-                + "timestamp DATE DEFAULT (datetime('now','localtime'))) "
+                + "timestamp DATE DEFAULT (datetime('now','localtime')),"
+                + "source_name TEXT NOT NULL)"
             )
 
             cur.execute(
                 "CREATE TABLE Volumes("
-                + "id INT,"
+                + "id INT NOT NULL,"
                 + "name TEXT,"
                 + "publisher TEXT,"
                 + "count_of_issues INT,"
                 + "start_year INT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "PRIMARY KEY (id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (id, source_name))"
             )
 
             cur.execute(
                 "CREATE TABLE AltCovers("
-                + "issue_id INT,"
+                + "issue_id INT NOT NULL,"
                 + "url_list TEXT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "PRIMARY KEY (issue_id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (issue_id, source_name))"
             )
 
             cur.execute(
                 "CREATE TABLE Issues("
-                + "id INT,"
+                + "id INT NOT NULL,"
                 + "volume_id INT,"
                 + "name TEXT,"
                 + "issue_number TEXT,"
@@ -117,10 +120,11 @@ class ComicVineCacher:
                 + "site_detail_url TEXT,"
                 + "description TEXT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "PRIMARY KEY (id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (id, source_name))"
             )
 
-    def add_search_results(self, search_term: str, cv_search_results: list[CVVolumeResults]) -> None:
+    def add_search_results(self, source_name: str, search_term: str, cv_search_results: list[CVVolumeResults]) -> None:
 
         con = lite.connect(self.db_file)
 
@@ -146,9 +150,10 @@ class ComicVineCacher:
 
                 cur.execute(
                     "INSERT INTO VolumeSearchCache "
-                    + "(search_term, id, name, start_year, publisher, count_of_issues, image_url, description) "
-                    + "VALUES(?, ?, ?, ?, ?, ?, ?, ?)",
+                    + "(source_name, search_term, id, name, start_year, publisher, count_of_issues, image_url, description) "
+                    + "VALUES(?, ?, ?, ?, ?, ?, ?, ?, ?)",
                     (
+                        source_name,
                         search_term.casefold(),
                         record["id"],
                         record["name"],
@@ -160,7 +165,7 @@ class ComicVineCacher:
                     ),
                 )
 
-    def get_search_results(self, search_term: str) -> list[CVVolumeResults]:
+    def get_search_results(self, source_name: str, search_term: str) -> list[CVVolumeResults]:
 
         results = []
         con = lite.connect(self.db_file)
@@ -173,7 +178,7 @@ class ComicVineCacher:
             cur.execute("DELETE FROM VolumeSearchCache WHERE timestamp  < ?", [str(a_day_ago)])
 
             # fetch
-            cur.execute("SELECT * FROM VolumeSearchCache WHERE search_term=?", [search_term.casefold()])
+            cur.execute("SELECT * FROM VolumeSearchCache WHERE search_term=? AND source_name=?", [search_term.casefold(), source_name])
             rows = cur.fetchall()
             # now process the results
             for record in rows:
@@ -186,6 +191,7 @@ class ComicVineCacher:
                         "description": record[7],
                         "publisher": {"name": record[4]},
                         "image": {"super_url": record[6]},
+                        "source": record[9],
                     }
                 )
 
@@ -193,7 +199,7 @@ class ComicVineCacher:
 
         return results
 
-    def add_alt_covers(self, issue_id: int, url_list: list[str]) -> None:
+    def add_alt_covers(self, source_name: str, issue_id: int, url_list: list[str]) -> None:
 
         con = lite.connect(self.db_file)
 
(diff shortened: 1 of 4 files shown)


comictaggerlib/comiccacher.py
```diff
@@ -282,7 +290,7 @@ def add_volume_issues_info(self, volume_id: int, cv_volume_issues: list[CVIssues
                     "description": issue["description"],
                     "timestamp": timestamp,
                 }
-                self.upsert(cur, "issues", "id", issue["id"], data)
+                self.upsert(cur, "issues", "id", issue["id"], data) # Don't need source?
```
lordwelch: upsert will not work as we need it it to, as it assumes that the value passed `issue["ID"]` will be a unique value, we currently do not have another source but that may change.

comictaggerlib/comiccacher.py
```diff
@@ -174,27 +179,28 @@ def get_search_results(self, search_term: str) -> list[CVVolumeResults]:
...
-                        "publisher": {"name": record[4]},
-                        "image": {"super_url": record[6]},
+                        "source": record[0],
+                        "id": record[2],
+                        "name": record[3],
+                        "start_year": record[4],
+                        "count_of_issues": record[6],
+                        "description": record[8],
+                        "publisher": {"name": record[5]},
+                        "image": {"super_url": record[7]},
```
lordwelch: I would recommend putting source as the last column it makes this block a one line change

comictaggerlib/comiccacher.py
```diff
@@ -298,7 +306,7 @@ def get_volume_info(self, volume_id: int) -> CVVolumeResults | None:
             cur.execute("DELETE FROM Volumes WHERE timestamp  < ?", [str(a_week_ago)])
 
             # fetch
-            cur.execute("SELECT id,name,publisher,count_of_issues,start_year FROM Volumes WHERE id = ?", [volume_id])
+            cur.execute("SELECT source_name,id,name,publisher,count_of_issues,start_year FROM Volumes WHERE id = ?", [volume_id]) # Don't need to where soure?
```
lordwelch: We do need to add the source to the condition.
right now if we were to add myanimelist through jikan
id: 4003 is https://myanimelist.net/manga/4003/Inviting_Darkness
on comic vine 
id: 4003 is https://comicvine.gamespot.com/green-arrow/4050-4003/

comictaggerlib/comiccacher.py
```diff
@@ -335,7 +344,7 @@ def get_volume_issues_info(self, volume_id: int) -> list[CVIssuesResults]:
 
             cur.execute(
                 (
-                    "SELECT id,name,issue_number,site_detail_url,cover_date,super_url,thumb_url,description"
+                    "SELECT source_name,id,name,issue_number,site_detail_url,cover_date,super_url,thumb_url,description"
                     " FROM Issues WHERE volume_id = ?"
```
lordwelch: This also needs to filter on source

comictaggerlib/comiccacher.py
```diff
@@ -346,13 +355,14 @@ def get_volume_issues_info(self, volume_id: int) -> list[CVIssuesResults]:
...
-                        "image": {"super_url": row[5], "thumb_url": row[6]},
-                        "description": row[7],
+                        "source_name": row[0],
+                        "id": row[1],
+                        "name": row[2],
+                        "issue_number": row[3],
+                        "site_detail_url": row[4],
+                        "cover_date": row[5],
+                        "image": {"super_url": row[6], "thumb_url": row[7]},
+                        "description": row[8],
```
lordwelch: Same as above I recommend making source the last column, as it makes this a one line change.

comictaggerlib/comiccacher.py
```diff
@@ -203,13 +209,13 @@ def add_alt_covers(self, issue_id: int, url_list: list[str]) -> None:
             cur = con.cursor()
 
             # remove all previous entries with this search term
-            cur.execute("DELETE FROM AltCovers WHERE issue_id = ?", [issue_id])
+            cur.execute("DELETE FROM AltCovers WHERE issue_id=? AND source_name=?", [issue_id], source_name)
```
lordwelch: wrong syntax should be [issue_id, source_name]

comictaggerlib/comicvinetalker.py
```diff
@@ -714,13 +717,13 @@ def parse_out_alt_cover_urls(self, page_html: str) -> list[str]:
...
-        url_list = cvc.get_alt_covers(issue_id)
+        cvc = ComicCacher()
+        url_list = cvc.get_alt_covers(self.source_name, issue_id)
 
         return url_list
 
     def cache_alternate_cover_urls(self, issue_id: int, url_list: list[str]) -> None:
-        cvc = ComicVineCacher()
+        cvc = ComicCacher()
         cvc.add_alt_covers(issue_id, url_list)
```
lordwelch: the source_name parameter is missing in this call

comictaggerlib/resulttypes.py
```diff
@@ -85,6 +86,7 @@ class CVVolume(TypedDict):
 
 
 class CVIssuesResults(TypedDict):
+    source: str
```
lordwelch: These CV result classes exist to model the json response from comicvine so should not have the source attribute added.

comictaggerlib/resulttypes.py
```diff
@@ -6,6 +6,7 @@
 
 
 class IssueResult(TypedDict):
+    source: str
```
lordwelch: This attribute is never set nor is it used

lordwelch (changes requested):

comictaggerlib/comiccacher.py
mizaki:
```suggestion
                if 'source_name' in keys:
                  condition += ' AND source_name = ?'
                  vals.append(data['source_name'])
```
Adding that between:
```
        sql_ins = f"INSERT OR IGNORE INTO {tablename} ({keys})  VALUES ({ins_slots})"
        cur.execute(sql_ins, vals)

        sql_upd = f"UPDATE {tablename} SET {set_slots} WHERE {condition}"
        cur.execute(sql_upd, vals)
```
Is a fix but I think it'd be better to alter:
```
def upsert(self, cur: lite.Cursor, tablename: str, pkname: str, pkval: Any, data: dict[str, Any]) -> None:
```
to
```
def upsert(self, cur: lite.Cursor, tablename: str, condition: dict[str, Any], data: dict[str, Any]) -> None:
```
Condition: {'source_name': 'comicvine', 'id': 001}
Does that seems like a better idea?

mizaki: Thinking about this a bit more, the better solution might be to create the PK from the source name and issue/volume id: "cv_001" or similar. I think the DB will always be small enough that performance of a text PK shouldn't be an issue. Or, give each source a number and combine that.

I think this seems better all around?

lordwelch: Altering the upsert function to take multiple keys is a good idea, creating a new primary key is even better (since we don't need to keep compatibility).
However sqlite lets us do this automatically using [insert or replace] and using multiple primary keys (I just looked this up)

[insert or replace]: https://www.sqlite.org/lang_replace.html

```patch
diff --git a/comictaggerlib/comiccacher.py b/comictaggerlib/comiccacher.py
index 7216279..6e1522d 100644
--- a/comictaggerlib/comiccacher.py
+++ b/comictaggerlib/comiccacher.py
@@ -90,28 +90,28 @@ class ComicCacher:
 
             cur.execute(
                 "CREATE TABLE Volumes("
-                + "id INT,"
+                + "id INT NOT NULL,"
                 + "name TEXT,"
                 + "publisher TEXT,"
                 + "count_of_issues INT,"
                 + "start_year INT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "source_name TEXT,"
-                + "PRIMARY KEY (id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (id, source_name))"
             )
 
             cur.execute(
                 "CREATE TABLE AltCovers("
-                + "issue_id INT,"
+                + "issue_id INT NOT NULL,"
                 + "url_list TEXT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "source_name TEXT,"
-                + "PRIMARY KEY (issue_id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (issue_id, source_name))"
             )
 
             cur.execute(
                 "CREATE TABLE Issues("
-                + "id INT,"
+                + "id INT NOT NULL,"
                 + "volume_id INT,"
                 + "name TEXT,"
                 + "issue_number TEXT,"
@@ -121,8 +121,8 @@ class ComicCacher:
                 + "site_detail_url TEXT,"
                 + "description TEXT,"
                 + "timestamp DATE DEFAULT (datetime('now','localtime')), "
-                + "source_name TEXT,"
-                + "PRIMARY KEY (id))"
+                + "source_name TEXT NOT NULL,"
+                + "PRIMARY KEY (id, source_name))"
             )
 
     def add_search_results(self, source_name: str, search_term: str, cv_search_results: list[CVVolumeResults]) -> None:
@@ -257,6 +257,7 @@ class ComicCacher:
                 pub_name = cv_volume_record["publisher"]["name"]
 
             data = {
+                "id": cv_volume_record["id"],
                 "source_name": source_name,
                 "name": cv_volume_record["name"],
                 "publisher": pub_name,
@@ -264,7 +265,7 @@ class ComicCacher:
                 "start_year": cv_volume_record["start_year"],
                 "timestamp": timestamp,
             }
-            self.upsert(cur, "volumes", "id", cv_volume_record["id"], data) # Don't need source?
+            self.upsert(cur, "volumes", data)
 
     def add_volume_issues_info(self, source_name: str, volume_id: int, cv_volume_issues: list[CVIssuesResults]) -> None:
 
@@ -279,6 +280,7 @@ class ComicCacher:
 
             for issue in cv_volume_issues:
                 data = {
+                    "id": issue["id"],
                     "volume_id": volume_id,
                     "source_name": source_name,
                     "name": issue["name"],
@@ -290,7 +292,7 @@ class ComicCacher:
                     "description": issue["description"],
                     "timestamp": timestamp,
                 }
-                self.upsert(cur, "issues", "id", issue["id"], data) # Don't need source?
+                self.upsert(cur, "issues", data)
 
     def get_volume_info(self, volume_id: int) -> CVVolumeResults | None:
 
@@ -382,13 +384,14 @@ class ComicCacher:
             timestamp = datetime.datetime.now()
 
             data = {
+                "id": issue_id,
                 "super_url": image_url,
                 "thumb_url": thumb_image_url,
                 "cover_date": cover_date,
                 "site_detail_url": site_detail_url,
                 "timestamp": timestamp,
             }
-            self.upsert(cur, "issues", "id", issue_id, data)
+            self.upsert(cur, "issues", data)
 
     def get_issue_select_details(self, issue_id: int) -> SelectDetails:
 
@@ -416,11 +419,10 @@ class ComicCacher:
 
             return details
 
-    def upsert(self, cur: lite.Cursor, tablename: str, pkname: str, pkval: Any, data: dict[str, Any]) -> None:
-        """This does an insert if the given PK doesn't exist, and an
-        update it if does
+    def upsert(self, cur: lite.Cursor, tablename: str, data: dict[str, Any]) -> None:
+        """This does an insert if the PK doesn't exist, and a
+        replace it if does
 
-        TODO: look into checking if UPDATE is needed
         TODO: should the cursor be created here, and not up the stack?
         """
 
@@ -443,13 +445,5 @@ class ComicCacher:
             ins_slots += "?"
             set_slots += key + " = ?"
 
-        keys += ", " + pkname
-        vals.append(pkval)
-        ins_slots += ", ?"
-        condition = pkname + " = ?"
-
-        sql_ins = f"INSERT OR IGNORE INTO {tablename} ({keys})  VALUES ({ins_slots})"
+        sql_ins = f"INSERT OR REPLACE INTO {tablename} ({keys}) VALUES ({ins_slots})"
         cur.execute(sql_ins, vals)
-
-        sql_upd = f"UPDATE {tablename} SET {set_slots} WHERE {condition}"
-        cur.execute(sql_upd, vals)
```

mizaki:
I think I've fixed all the issues. -What's the best way to proceed: Close this and fix the merge issues and then make a new PR?-

Okay, I think I worked it out.

lordwelch:
First delete your last commit a1e656a it wasn't merged properly, then do a [rebase] and then [squash] your commits.
Once that's done just run 'git push --force'

[rebase]: https://stackoverflow.com/a/18442755
[squash]: https://stackoverflow.com/a/5189600

mizaki:
I think that was successful?

comictaggerlib/comiccacher.py
```diff
@@ -334,24 +345,25 @@ def get_volume_issues_info(self, volume_id: int) -> list[CVIssuesResults]:
 
             cur.execute(
                 (
-                    "SELECT id,name,issue_number,site_detail_url,cover_date,super_url,thumb_url,description"
-                    " FROM Issues WHERE volume_id = ?"
+                    "SELECT source_name,id,name,issue_number,site_detail_url,cover_date,super_url,thumb_url,description"
+                    " FROM Issues WHERE id=? AND source_name=?"
```
lordwelch: id needs to be reverted to volume_id

lordwelch:
Two things
* you don't need to merge the develop branch into your branch it just makes the commit history confusing because the change that it added was about image resizing which has no relation to the cache db
* there is an issue on [line 349] with the volume_id

[line 349]: https://github.com/comictagger/comictagger/pull/323/files#r914039865

mizaki:
Damn, I thought I removed the popup window change. Apologies.

lordwelch (approved):

PR #363, opened 2022-08-25
re-work settings

lordwelch:
My goal is to combine settings and cli parameters as much as possible and use JSON, TOML or another format that is more structured.


comictaggerlib/main.py
```diff
@@ -15,226 +15,144 @@
...
+        self.register_options()
+        self.parse_options(opts.config)
+        self.initialize_dirs()
+
+        self.ctmain()
+
+    def initialize(self) -> argparse.Namespace:
+        opts, _ = self.initial_arg_parser.parse_known_args()
+        assert opts is not None
+        setup_logging(opts.verbose, opts.config.user_log_dir / "log")
```
mizaki: The `/ "log"` is causing the /log/log issue.

comictaggerlib/settings/cmdline.py
```diff
@@ -0,0 +1,343 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+def initial_cmd_line_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(add_help=False)
+    # Ensure this stays up to date with register_options
+    parser.add_argument(
+        "--config",
+        help="Config directory defaults to ~/.ComicTagger\non Linux/Mac and %%APPDATA%% on Windows\n",
```
mizaki: Linux is ~/.config/ComicTagger

comictaggerlib/main.py
lordwelch: Done

comictaggerlib/settings/cmdline.py
lordwelch: Done

PR #373, opened 2022-09-16
Modified rename to allow file path to be used in rename.

rhaussmann:
This allows OSX to use file paths during renaming. `**self**.move` boolean seems to be True even when it shows False on the logging, so it was removed. `pathlib.Path` should use the path format of the native system.

### DIFF
--- a/comictaggerlib/filerenamer.py
+++ b/comictaggerlib/filerenamer.py
@@ -37,7 +37,7 @@ class Replacements(NamedTuple):
 
 
 REPLACEMENTS = Replacements(
-    literal_text=[(": ", " - "), (":", "-")], format_value=[(": ", " - "), (":", "-"), ("/", "-"), ("\\", "-")]
+    literal_text=[(": ", " - "), (":", "-")], format_value=[(": ", " - "), (":", "-")]
 )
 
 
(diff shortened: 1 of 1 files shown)


lordwelch:
@rhaussmann please select the `Move files when renaming` check box to accomplish what you need, If the destination directory is not filled out it will assume the current working directory which for a GUI app on macOS is the root, which will currently cause a crash. Please set it to the appropriate path and you should have no issues. I have just tested this on macOS with the latest release as posted.
Screen Shot 2022-09-15 at 8 03 49 PM

As your main issue is already a supported feature, switching to pathlib.Path instead of of pathlib.PureWindowsPath will not be accepted as it allows easier management of cross-platform strings representing filepaths the only major effect it has is related to slashes of which it only effects the text that you type into the template textbox directly it has no effect on any text produced from the metadata. For example the template `{publisher}\{title}` or `{publisher}/{title}` for 'batman/superman' will end up as `batmansuperman` if strict renaming is not selected (this is a bug it should show the same as the next) other wise it will show as `batman-superman` if you further select the `Move files when renaming` check box it will show as `DC/batmansuperman` or `DC/batman-superman`. Your change to the replacements is not warranted and will not be accepted as they currently only happen when platform is set to windows or universal which corresponds to the strict renaming checkbox.

rhaussmann:
I appreciate the detailed response! 
I just got it to work using your direction. Thanks!

 I just never tried them together while trying to build directories. I think I was put off by the **move files when renaming** because I didn't want to move to another directory. Thanks again to you all for your hard work.

PR #403, opened 2022-11-28
Talkers part 2

mizaki:


mizaki:
Do you have another preferred way to achieve the separation of talkers with issues and without?

mizaki:
Is https://github.com/comictagger/comictagger/pull/403/commits/66418877f8bebe6e16d86733ad9fab501bb323ae the general idea you had with using the settings in the talker?

Is the idea to also generate the settings menus from the `options["definitions"][<talker>]` data?

comictalker/talkers/comicvine.py
```diff
@@ -163,10 +165,74 @@ class CVIssueDetailResults(TypedDict, total=False):
...
+    )
+
+
 class ComicVineTalker(ComicTalker):
     default_api_key = "27431e6787042105bd3e47e169a624521f89f3a4"
     default_api_url = "https://comicvine.gamespot.com/api"
 
+    @staticmethod
+    def register_settings(parser: settngs.Manager) -> None:
+        parser.add_group("comicvine", comicvine_settings, False)
```
lordwelch: This is almost right. The Talker will not be calling add_group, ComicTagger will call add_group with this function.
Also let's not use `staticmethod` that will just cause further complications.
Let's remove all of the arguments in the constructor except for `version` and `cache_folder`, and make a function like `parse_settings` that will take a dictionary of the options they defined also about half of these options will stay with ComicTagger instead of being defined here

lordwelch (commented):
I would rather have MangaUpdates return a list of issues with the only difference being the issue number than trying to shove random volume data into IssueIdentifier. In IssueIdentifier instead of separating volume data from issue data I would rather just make it better at identifying issue data without a cover. Also I'm going to change everything in CT to use Series instead of Volume, it will make things easier to think about with manga.

lordwelch:
> Is the idea to also generate the settings menus from the `options["definitions"][<talker>]` data?

More or less, I already changed where exactly that is stored so there isn't a `options["definitions"]` anymore, I'll probably tweak what exactly it looks like before I publish the next version

mizaki:
> > Is the idea to also generate the settings menus from the `options["definitions"][<talker>]` data?
> 
> More or less, I already changed where exactly that is stored so there isn't a `options["definitions"]` anymore, I'll probably tweak what exactly it looks like before I publish the next version

To my mind it would be ideal if the "definition" and the values were in a dict named after the "section". Maybe that's the plan? That way the settings menu could be generated and values set from that one dict. Also being able to send that to the talker which is what I think you are also saying about keeping only `version` and `cache_folder`?

comictalker/talkers/comicvine.py
mizaki: I did consider some of the options being general, as I did last but I think it makes sense to keep them unique to each talker. With one talker you may want to sort by year but not another for example. Even with the publisher filter. Which of the settings do you think should be general to all talkers?

lordwelch: Never referenced in comictaggerlib except in relation to setting the value:
* **--series-match-search-thresh**
* **--use-series-start-as-volume**
* **--remove-html-tables**
* **--cv-api-key**
* **--cv-url**

Only referenced in comictaggerlib:
* **--clear-metadata**: Does not make sense to send to the Talker, the Talker never sees any metadata to override
* **--auto-imprint**: Could be used but now depends on loading external data and if a Talker does not implement it it won't be there, leaving it in comictaggerlib ensures that it can always be used
* **--sort-series-by-year**: Makes sense as a potential option but if a Talker does not implement it it won't be there, leaving it in comictaggerlib ensures that it can always be used. Instead of moving this to the Talker I would rather have an option to have comictaggerlib not sort the returned data
* **--exact-series-matches-first**: Same as above
* **--always-use-publisher-filter**: Could be used but if a Talker does not implement it it won't be there, leaving it in comictaggerlib ensures that it can always be used
* **--clear-form-before-populating-from-cv**: Does not make sense to send to the Talker, it is a library and does not interact with any forms

It would make sense to make it so that some of these can be set per Talker but that would add additional complexity that I do not want to deal with. Primarily it would duplicate the settings to each talker, which in turn would mean that you could tune precisely what happens for several different sources, but what happens when a new source is added, what settings should we use? What would a user that just downloaded ComicTagger and used ComicVine, the default source, and configured it to do what they want expect when they add a new source? For the technically savvy, they are probably using it in a script which means that for their primary usage they can update their script to use different command-line parameters for each source, and if it's a file only option, which there aren't many left, they can use the --config option to point to a different folder. Yes it means that it is not a great experience if they want a different filter for each source, but there is also nothing stopping a source from implementing their own filter option, e.g. `--mu-publisher-filter`, the only real restriction on what types can be used, is that if it is a file settings it must be a type that the default json [encoder] and [decoder] can process.

Also note that since comictalker is a separate library I intend as much as possible for nothing to be added to it that would force it to be used with ComicTagger it should be something that can be extracted and used in a separate project (we'll see how well this actually turns out)

[decoder]: https://docs.python.org/3/library/json.html#json.JSONDecoder
[encoder]: https://docs.python.org/3/library/json.html#json.JSONEncoder

lordwelch:
Currently the definitions are returned as a tuple specifically this type https://github.com/lordwelch/settngs/blob/c3976c2fb5a3be78d7c0ac322e5f7465eec2b257/settngs.py#L177
I don't see why a talker would ever need to see the definition of it's options, the primary reason for even returning it from the Manager class is so that settings do not need to be registered again to save the settings or do another operation with a newly created Manager object

comictalker/talkers/comicvine.py
mizaki: I've split them out in https://github.com/comictagger/comictagger/pull/403/commits/d36337edbcafcde3b7f722563485915349d81beb may still want to move some around.

I now have the `talkerapi` adding the group settings after being called from `main` `register_options` but that may not be what you intend either?

With the constructor and settings is the idea to pass in `options[0]` and then have the talker parse out it's own settings?

lordwelch: No, like [flake8], there would be a parse_options function that I would send a dictionary with all of the plugin's options.
As an example this is what would be currently sent to ComicVineTalker on d36337edbcafcde3b7f722563485915349d81beb
```python
{
    "series_match_search_thresh": 90,
    "use_series_start_as_volume": False,
    "remove_html_tables": False,
    "cv_api_key": "",
    "cv_url": "",
}
```

[flake8]: https://flake8.pycqa.org/en/latest/plugin-development/plugin-parameters.html#accessing-parsed-options

mizaki:
> I would rather have MangaUpdates return a list of issues with the only difference being the issue number than trying to shove random volume data into IssueIdentifier. In IssueIdentifier instead of separating volume data from issue data I would rather just make it better at identifying issue data without a cover. Also I'm going to change everything in CT to use Series instead of Volume, it will make things easier to think about with manga.

With https://github.com/comictagger/comictagger/pull/403/commits/f61f7ce803143496034d031e5b7c549d51c506a4 I've made minimal changes to issue identifier to enable a series only talker to use it. As the problem arises when ii wants a list of issues, I've copied over the available series data to create the issue list. This seems like a better idea than creating a "fake" `fetch_issues_by_volume_issue_num_and_year` function to return the same data possibly with an additional year filter.

The only thing wanted really is filtering down the results for the cover match.

Would you prefer the "fake" `fetch_issues_by_volume_issue_num_and_year` instead? Or some other way (that I can't think of right now)?

lordwelch:
@mizaki you are making this too complicated see #412 and 9f00af4bba0a3410bdce05a5ef54fcc4d48be7c0 for the minimum required changes to make MangaUpdates work. Note that it is not production ready by any means but it does work.

mizaki:
> @mizaki you are making this too complicated see #412 and [9f00af4](https://github.com/comictagger/comictagger/commit/9f00af4bba0a3410bdce05a5ef54fcc4d48be7c0) for the minimum required changes to make MangaUpdates work. Note that it is not production ready by any means but it does work.

As I see it that requires an issue number to work even if the requirement is removed from `ii`? No issue number will be available (outside of the file name may have a chapter etc. number but what to do about those is a different conversation). 

Maybe `ii` isn't really what's needed as it's only being used to compare the cover? The search has already need done. It looks like the cover comparison is tied up in the `ii` `search` function. Could `get_issue_cover_match_score` be used alone or add a function to compare two images?

I may be causing confusion by using the nomenclature "issue". To fit within the framework that's already there I was using "series" for the search series data and "issue" as the full series data (for caching etc.).

Re: #412 
`fetch_issues_by_series` isn't going to be called by a series only talker as the volume window disables opening the issues window. Even if it did, all the window would show is the same as the volume window but with only one item.

`fetch_issues_by_series_issue_num_and_year` ignoring the issue number requirement, this will only ever return a single item which is what I meant by a "fake" version. It will retrieve more information than that passed in by the search result but none of that extra information will be used. That's why I duplication it in my `ii` hack.

Hopefully all that makes sense and I'm not being dumb somewhere? :)

mizaki:
After some more thinking. Another maybe is to split the `ii` search function into `series` and `issues`. Right now the two are combined. `has_issues` could decide (or a flag passed in) if the series search function in `ii` should compare the covers and then pass the results on to the issue identifying part.

Might be making it more complicated again but it seems the cleanest to me right now.

mizaki:
I've had a look around flake8 and tried to do something similar in https://github.com/comictagger/comictagger/pull/403/commits/edf07ee4edfbf83c0dad90729e6a684263165ba0 for adding and parsing plugin settings. Anywhere near?

lordwelch:
@mizaki You have at least 4 different things, that I could find, are attempted in this pull request and none of them depend on each other:
* Remove the requirement of an issue number when searching
* Implement a fully operational plugin system for use with a ComicTalker
* Add MangaUpdates as a useable alternative to ComicVine
* Separate settings that are particular to a specific talker. e.g. ComicVine/MangaUpdates specific settings

 I've closed this pull request. Open a new PR and pick one thing to implement and base it off of the current develop branch (currently at commit 3311f3ae52e00c8ed6845bef98e97f5e4b02b1f9)

PR #404, opened 2022-11-28
Fix some keys logic in issue identifier

mizaki:
keys may be None if there is no comic archive. 

IssueString.as_string will convert None to empty string so use None comparison before.

### DIFF
--- a/comictaggerlib/issueidentifier.py
+++ b/comictaggerlib/issueidentifier.py
@@ -177,8 +177,6 @@ class IssueIdentifier:
 
         ca = self.comic_archive
 
-        if ca is None:
-            return None
         search_keys: SearchKeys
         if self.only_use_additional_meta_data:
             search_keys = SearchKeys(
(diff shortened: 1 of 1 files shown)


lordwelch:
ca is guaranteed to be not none by the time this executes see lines https://github.com/comictagger/comictagger/blob/fc4eb4f002a11ceb0a06e2eb78421d3b61ba5ae3/comictaggerlib/issueidentifier.py#L335 and https://github.com/comictagger/comictagger/blob/fc4eb4f002a11ceb0a06e2eb78421d3b61ba5ae3/comictaggerlib/issueidentifier.py#L339

This "issue" has existed since 2013 3072583482deacd12e5d8658cb835b44127adbdb so I doubt it actually causes any issues. In any case I would rather change it to:
```python
        # we need, at minimum, a series and issue number
        if not (keys["series"] and keys["issue_number"]):
            self.log_msg("Not enough info for a search!")
            return []
```

PR #405, opened 2022-11-28
Better resize columns in the VolumeSelectionWindow

### LINKED ISSUE #402: Publisher name gets cut-off when the series name is too long
It's mainly the search window that this becomes annoying with but others should probably remember their size and position too.

lordwelch:
Fixes #402

### DIFF
--- a/comictaggerlib/volumeselectionwindow.py
+++ b/comictaggerlib/volumeselectionwindow.py
@@ -479,7 +479,7 @@ class VolumeSelectionWindow(QtWidgets.QDialog):
         # Get the width of the issues, year and publisher columns
         owidth = self.twList.columnWidth(1) + self.twList.columnWidth(2) + self.twList.columnWidth(3)
         # Get the remaining width after they fill the tableWidget
-        rwidth = self.twList.width() - owidth
+        rwidth = self.twList.contentsRect().width() - owidth
 
         # Default the tableWidget to truncate series names
         self.twList.setColumnWidth(0, rwidth)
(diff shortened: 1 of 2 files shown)


comictaggerlib/volumeselectionwindow.py
```diff
@@ -474,6 +474,16 @@ def categorize(result: ComicVolume) -> int:
         self.twList.setSortingEnabled(True)
         self.twList.selectRow(0)
         self.twList.resizeColumnsToContents()
+        # Get the width of the issues, year and publisher columns
+        owidth = self.twList.columnWidth(1) + self.twList.columnWidth(2) + self.twList.columnWidth(3)
```
mizaki, line 480: There is a margin on the right which is lost on the publisher column. Adding about 15 seems to be about right:
```
owidth = self.twList.columnWidth(1) + self.twList.columnWidth(2) + (self.twList.columnWidth(3) + 15)
```

mizaki:
I would also argue for making the window a bit wider and taller. As it's modal leaving about 10% space on each side would allow for less likelihood of needing two lines for a title.

comictaggerlib/volumeselectionwindow.py
lordwelch: Fixed by using:
```python
self.twList.contentsRect().width() - owidth
```

mizaki:
Looks good. Window could still suffer being a little bit bigger though but I won't mention it.

PR #406, opened 2022-11-30
Sort volume window by fuzz ratio first

mizaki:
This is just to get going.

- If the ratio is to be cached it probably belongs in the `VolumeSearchCache` table.
- If the talker doesn't fetch the ratio, `volumeselectionwindow` could do it (and ratio wouldn't need to be cached).

As far as I see, `issueidentifier` doesn't get involved unless some form of auto-identify is involved.

### DIFF
--- a/comicapi/utils.py
+++ b/comicapi/utils.py
@@ -162,10 +162,14 @@ def sanitize_title(text: str, basic: bool = False) -> str:
     return text
 
 
-def titles_match(search_title: str, record_title: str, threshold: int = 90) -> bool:
-    sanitized_search = sanitize_title(search_title)
-    sanitized_record = sanitize_title(record_title)
-    ratio: int = rapidfuzz.fuzz.ratio(sanitized_search, sanitized_record)
+def titles_match(search_title: str, record_title: str, threshold: int = 90, basic: bool = True) -> tuple[bool, float]:
+    if basic:
+        sanitized_search = sanitize_title(search_title)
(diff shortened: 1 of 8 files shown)


lordwelch:
1. This is way more complicated than it needs to be: [Patch.txt]
1. There is no useful  nor practical reason to cache this information
   1. Clear your cache, run a search on 'batman', run a search on 'batman/superman', and then go back and do another search on 'batman' and you'll see the issue
1. [Here] you did the equivalent of:
   ```python
   if value == True:
       return False
   else:
       return True
   ```
   and you incorrectly reversed the meaning of basic
   This is what it should be if you were planning on adding the basic argument to titles_match
   ```python
   sanitized_search = sanitize_title(search_title, basic)
   ```
1. In the same function as above you change what happens by default which now returns a different set of results: at a minimum when searching 'batman' previously [Batman and...] would show up directly after the 20+ series of batman, now it isn't even a result
1. The current sort in ComicTagger sorts by year and number of issue because most of the time those are most likely the series you are trying to tag, it then prioritizes any issues that exactly or closely match the original search string, which provides a better experience than essentially sorting by series name length

[Patch.txt]: https://github.com/comictagger/comictagger/files/10118778/Patch.txt
[Batman and...]: https://comicvine.gamespot.com/batman-and/4050-87345/
[Here]: https://github.com/mizaki/comictagger/blob/ed95c9ee53b97d30c11a406bc9d7dfaf8da26f4d/comicapi/utils.py#L166-L171

PR #418, opened 2023-01-01
Use entry points for talker plugins.

mizaki:
May wish to add more to the plugin list (load() the ep for a friendly display name etc. from a class var). 

Will probably need to pass the list into the settings window so creating a self.plugins from `get_talkers` in main.

### DIFF
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -191,7 +191,4 @@ class App:
             except Exception:
                 logger.exception("CLI mode failed")
         else:
-            if error and error[1]:
-                print(f"A fatal error occurred please check the log for more information: {error[0]}")  # noqa: T201
-                raise SystemExit(1)
             gui.open_tagger_window(talker_api, self.options, error)
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -150,7 +150,7 @@ class App:
                 return
 
         try:
-            talker_api = self.talker_plugins["comicvine"](  # type: ignore[operator, call-arg]
+            talker_api = self.talker_plugins["comicvine"](  # type: ignore[call-arg]
                 version=version,
                 cache_folder=self.options[0].runtime_config.user_cache_dir,
                 series_match_thresh=self.options[0].comicvine_series_match_search_thresh,
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -66,6 +66,7 @@ class App:
         self.options = settngs.Config({}, {})
         self.initial_arg_parser = ctoptions.initial_cmd_line_parser()
         self.config_load_success = False
+        self.talker_plugins = ct_api.get_talkers()
 
     def run(self) -> None:
         opts = self.initialize()
@@ -149,7 +150,7 @@ class App:
                 return
 
         try:
-            talker_api = ct_api.get_comic_talker("comicvine")(  # type: ignore[call-arg]
+            talker_api = self.talker_plugins["comicvine"](  # type: ignore[operator, call-arg]
                 version=version,
                 cache_folder=self.options[0].runtime_config.user_cache_dir,
                 series_match_thresh=self.options[0].comicvine_series_match_search_thresh,
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -112,6 +112,22 @@ class App:
         logger.debug("user_log_dir: %s", self.options[0].runtime_config.user_log_dir)
 
     def main(self) -> None:
+        def load_talker(talker):
+            try:
+                return self.talker_plugins[talker](  # type: ignore[call-arg]
+                    version=version,
+                    cache_folder=self.options[0].runtime_config.user_cache_dir,
+                    series_match_thresh=self.options[0].comicvine_series_match_search_thresh,
+                    remove_html_tables=self.options[0].comicvine_remove_html_tables,
+                    use_series_start_as_volume=self.options[0].comicvine_use_series_start_as_volume,
+                    wait_on_ratelimit=self.options[0].autotag_wait_and_retry_on_rate_limit,
+                    api_url=self.options[0].comicvine_cv_url,
+                    api_key=self.options[0].comicvine_cv_api_key,
+                )
+            except Exception:
+                logger.exception("Unable to load talker %s", talker)
+                return False
+
         assert self.options is not None
         # options already loaded
         error = None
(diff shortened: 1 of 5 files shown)


lordwelch:
So right now if `get_comic_talker` is called on start to get 'comicvine' and then it is called again to load 'mangaupdates' it goes back to start and re-parses and re-loads every module it needs to load, it would be better to return a class/tuple/list/etc that contains all of the plugins already loaded

mizaki:
> So right now if `get_comic_talker` is called on start to get 'comicvine' and then it is called again to load 'mangaupdates' it goes back to start and re-parses and re-loads every module it needs to load, it would be better to return a class/tuple/list/etc that contains all of the plugins already loaded

As there is only one talker atm I didn't want to muddy the waters. https://github.com/comictagger/comictagger/pull/419 does load everything into a `self.plugins` (although it still has the problem mentioned). I was trying to keep things separate and so all this is really, is adding the ability to load plugins through entry points.

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,50 @@
...
 
-def get_comic_talker(source_name: str) -> type[ComicTalker]:
-    """Retrieve the available sources modules"""
-    sources = get_talkers()
-    if source_name not in sources:
-        raise TalkerError(source=source_name, code=4, desc="The talker does not exist")
+class Plugin(NamedTuple):
+    package: str
+    version: str
+    entry_point: EntryPoint
```
lordwelch: package and version is not currently used and doesn't work well with what is assigned to it later.

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,50 @@
...
+
+def get_talkers() -> dict[str, Plugin]:
+    """Returns all comic talker plugins (internal and external)"""
+    talkers: dict[str, Plugin] = {}
+    for dist in distributions():
+        eps = dist.entry_points
+
+        # perf: skip parsing `.metadata` (slow) if no entry points match
+        if not any(ep.group in ("comictagger_talkers") for ep in eps):
+            continue
```
lordwelch: The only information pulled out of the `.metadata` is the version which isn't used so this can be removed.

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,50 @@
...
 
-    talker = sources[source_name]
-    return talker
+        # assigned to prevent continual reparsing
+        meta = dist.metadata
```
lordwelch: Same as above.

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,50 @@
...
 
+        for ep in eps:
+            # Only want comic talkers
+            if ep.group == "comictagger_talkers":
+                talkers[ep.name] = Plugin(ep.name, meta["version"], ep)
```
lordwelch: This should load the entry point we don't have a reason not to and it needs a try block to make sure that the entry point loads successfully
```suggestion
        for ep in eps:
            # Only want comic talkers
            if ep.group == "comictagger_talkers":
                try:
                    talkers[ep.name] = ep.load()
                except Exception:
                    logger.exception("Failed to load talker: %s", ep.name)
```

lordwelch (commented):
In addition to the other comments there is an `entry_points` function that importlib.metadata provides to get only the entry points so we can get rid of one of the loops and as one of many ways to do it, we can import `cache` from functools to cache the list of talkers, and we are left with this.

```python
from functools import cache

if sys.version_info < (3, 10):
    from importlib_metadata import entry_points
else:
    from importlib.metadata import entry_points

@cache
def get_talkers():
    """Returns all comic talker plugins (internal and external)"""
    talkers = {}

    for ep in entry_points(group="comictagger_talkers"):
        try:
            talkers[ep.name] = ep.load()
        except Exception:
            logger.exception("Failed to load talker: %s", ep.name)

    return talkers
```

comictalker/comictalkerapi.py
mizaki: My thinking here is the version of the external plugins should be logged as CT version and talker plugins will be unbound. If not adding it to `Plugin` class then maybe a simple debug in `get_talkers` is a better idea? Probably want to put in a `logger.info` for the plugins found anyway? An alternative could be a class var (which seems to be needed for a display name, i.e. Comic Vine)?

comictalker/comictalkerapi.py
mizaki: That is basically what I had originally but I wanted to get the version number in and followed flake8. I don't think the cache is needed if, for now, there is only one talker and when the option to switch talkers will be in another PR. In that PR, I don't think `get_comic_talker` will be needed. Do you want me to add it in now anyway?

comictalker/comictalkerapi.py
mizaki: I see that externals are already taken care of via `main` debug logging.

comictalker/comictalkerapi.py
lordwelch: If you don't want to cache it, remove the function for retrieving an individual Talker. It's a glorified dict lookup.

comictalker/comictalkerapi.py
lordwelch: For now lets keep this all internal, anything relating to external talkers should happen later.

comictalker/comictalkerapi.py
mizaki: Removed `cache` and `get_comic_talker`. `main` calls `get_talkers` and stores it in `self.talker_plugins` now on `init`.

mizaki:
I'm not sure on the best way to handle possible errors when loading the talker, see https://github.com/comictagger/comictagger/pull/418/commits/56577f7cea871317bf35c5841297c6113ff21a82

I'm wonder if having a loader in `talkerapi` might be the better way and return the object. Something like there was before:
```
def get_comic_talker(source_name: str, plugins: dict[str, type[ComicTalker], params: dict[str, Any]) -> ComicTalker:
    """Returns talker object"""
    if source_name not in plugins:
        raise TalkerError(source=source_name, code=4, desc="The talker does not exist")

    try:
        return plugins[source_name](params)
    except:
        raise TalkerError(source=source_name, code=4, desc="The talker failed to load")
```
And in `main`:
```
     try:
        talker_api = ct_api.get_comic_talker(all_the_params)
```
Would that be a better approach? I'm also thinking ahead for when switching talkers will be required, having a function to call would be better.

lordwelch:
Error handling ends up being about the same no matter what.
From what I see in effect you are just wrapping the exception for you to catch it again. I don't see any change in the use of it.
In main.py moving all of the Talker loading into another function would be nice eg `def load_talker(self):`

mizaki:
The `load_talker` being in `comictalkerapi` makes more sense later and other things will change too so it can re-evaluated then. Put it in `main` for now.

comictaggerlib/main.py
```diff
@@ -148,20 +164,18 @@ def main(self) -> None:
...
-        except TalkerError as e:
-            logger.exception("Unable to load talker")
-            error = (str(e), True)
+        talker_api = load_talker("comicvine")
+        if not talker_api:
+            if self.options[0].runtime_no_gui:
+                # Need to hard exit on talker error with the CLI
+                error = (
+                    "Unable to load talker. Please check your source name ID is correct in your config file or your command line option.",
+                    True,
```
lordwelch, line 173: Move all of this and `def load_talker(talker):` to it's own method. Directly return the error

comictaggerlib/main.py
```diff
@@ -177,4 +191,7 @@ def main(self) -> None:
             except Exception:
                 logger.exception("CLI mode failed")
         else:
+            if error and error[1]:
+                print(f"A fatal error occurred please check the log for more information: {error[0]}")  # noqa: T201
+                raise SystemExit(1)
```
lordwelch: This is incorrect please remove.
The error is handled in `gui.open_tagger_window(talker_api, self.options, error)`

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,32 @@
...
+logger = logging.getLogger(__name__)
 
 
 def get_talkers() -> dict[str, type[ComicTalker]]:
-    """Returns all comic talker modules NOT objects"""
-    return {"comicvine": comictalker.talkers.comicvine.ComicVineTalker}
+    """Returns all comic talker plugins (internal and external)"""
+    talkers = {}
+
+    for ep in entry_points(group="comictagger_talkers"):
```
lordwelch: This should be `comictagger.talkers` see https://packaging.python.org/en/latest/specifications/entry-points/#data-model

comictaggerlib/main.py
mizaki: As https://github.com/comictagger/comictagger/pull/419 (plugin settings) replaces this anyway, I'll leave this and make any additional changes required after the plugin settings is merged.

PR #419, opened 2023-01-02
Use new settings system for plugin

mizaki:
Currently this doesn't update the current talker object but otherwise how is this?

### DIFF
--- a/comictaggerlib/ctoptions/file.py
+++ b/comictaggerlib/ctoptions/file.py
@@ -96,7 +96,11 @@ def filename(parser: settngs.Manager) -> None:
 def talkers(parser: settngs.Manager) -> None:
     # General settings for all information talkers
     parser.add_setting("--source", default="comicvine", help="Use a specified source by source ID")
-    parser.add_setting("--use-series-start-as-volume", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting(
+        "--series-match-search-thresh",
+        default=90,
+        type=int,
+    )
     parser.add_setting(
         "--clear-metadata",
         default=True,
--- a/comictaggerlib/ctoptions/file.py
+++ b/comictaggerlib/ctoptions/file.py
@@ -93,13 +93,9 @@ def filename(parser: settngs.Manager) -> None:
     )
 
 
-def comicvine(parser: settngs.Manager) -> None:
-    # Comic Vine settings
-    parser.add_setting(
-        "--series-match-search-thresh",
-        default=90,
-        type=int,
-    )
+def talkers_general(parser: settngs.Manager) -> None:
+    # General settings for all information talkers
+    parser.add_setting("--source", default="comicvine", help="Use a specified source by source ID")
     parser.add_setting("--use-series-start-as-volume", default=False, action=argparse.BooleanOptionalAction)
     parser.add_setting(
         "--clear-metadata",
@@ -107,20 +103,6 @@ def comicvine(parser: settngs.Manager) -> None:
         help="Clears all existing metadata during import, default is to merges metadata.\nMay be used in conjunction with -o, -f and -m.\n\n",
         dest="clear_metadata_on_import",
         action=argparse.BooleanOptionalAction,
-    )
-    parser.add_setting(
-        "--remove-html-tables",
-        default=False,
-        action=argparse.BooleanOptionalAction,
-        help="Removes html tables instead of converting them to text",
-    )
-    parser.add_setting(
-        "--cv-api-key",
-        help="Use the given Comic Vine API Key (persisted in settings).",
-    )
-    parser.add_setting(
-        "--cv-url",
-        help="Use the given Comic Vine URL (persisted in settings).",
     )
     parser.add_setting(
         "-a",
@@ -146,10 +128,10 @@ def comicvine(parser: settngs.Manager) -> None:
         help="Enables the publisher filter",
     )
     parser.add_setting(
-        "--clear-form-before-populating-from-cv",
+        "--clear-form-before-populating",
         default=False,
         action=argparse.BooleanOptionalAction,
-        help="Clears all existing metadata when applying metadata from ComicVine",
+        help="Clears all existing metadata when applying metadata from comic source",
     )
 
 
@@ -257,13 +239,17 @@ def validate_settings(options: settngs.Config[settngs.Values], parser: settngs.M
     return options
 
 
-def register_settings(parser: settngs.Manager) -> None:
+def register_settings(parser: settngs.Manager, talkers: dict[str, Any]) -> None:
     parser.add_group("general", general, False)
     parser.add_group("internal", internal, False)
     parser.add_group("identifier", identifier, False)
     parser.add_group("dialog", dialog, False)
     parser.add_group("filename", filename, False)
-    parser.add_group("comicvine", comicvine, False)
+    parser.add_group("talkers_general", talkers_general, False)
     parser.add_group("cbl", cbl, False)
     parser.add_group("rename", rename, False)
     parser.add_group("autotag", autotag, False)
+
+    # Register talker plugin settings
+    for talker, cls in talkers.items():
+        parser.add_group(talker, cls.comic_settings, False)
--- a/comictaggerlib/ctoptions/file.py
+++ b/comictaggerlib/ctoptions/file.py
@@ -93,7 +93,7 @@ def filename(parser: settngs.Manager) -> None:
     )
 
 
-def talkers_general(parser: settngs.Manager) -> None:
+def talkers(parser: settngs.Manager) -> None:
     # General settings for all information talkers
     parser.add_setting("--source", default="comicvine", help="Use a specified source by source ID")
     parser.add_setting("--use-series-start-as-volume", default=False, action=argparse.BooleanOptionalAction)
@@ -239,17 +239,13 @@ def validate_settings(options: settngs.Config[settngs.Values], parser: settngs.M
     return options
 
 
-def register_settings(parser: settngs.Manager, talkers: dict[str, Any]) -> None:
+def register_settings(parser: settngs.Manager) -> None:
     parser.add_group("general", general, False)
     parser.add_group("internal", internal, False)
     parser.add_group("identifier", identifier, False)
     parser.add_group("dialog", dialog, False)
     parser.add_group("filename", filename, False)
-    parser.add_group("talkers_general", talkers_general, False)
+    parser.add_group("talkers", talkers, False)
     parser.add_group("cbl", cbl, False)
     parser.add_group("rename", rename, False)
     parser.add_group("autotag", autotag, False)
-
-    # Register talker plugin settings
-    for talker, cls in talkers.items():
-        parser.add_group(talker, cls.comic_settings, False)
--- a/comictaggerlib/ctoptions/talker_plugins.py
+++ b/comictaggerlib/ctoptions/talker_plugins.py
@@ -4,14 +4,14 @@ import logging
 
 import settngs
 
-from comictalker.comictalkerapi import TalkerPlugin
+from comictalker.talkerbase import ComicTalker
 
 logger = logging.getLogger(__name__)
 
 
-def register_talker_settings(parser: settngs.Manager, plugins: dict[str, TalkerPlugin]) -> None:
+def register_talker_settings(parser: settngs.Manager, plugins: dict[str, ComicTalker]) -> None:
     for talker_name, talker in plugins.items():
         try:
-            parser.add_group(talker_name, talker["cls"].comic_settings, False)
+            parser.add_group(talker_name, talker.comic_settings, False)
         except Exception:
             logger.exception("Failed to register settings for %s", talker_name)
--- a/comictaggerlib/ctoptions/talker_plugins.py
+++ b/comictaggerlib/ctoptions/talker_plugins.py
@@ -13,6 +13,6 @@ logger = logging.getLogger(__name__)
 def register_talker_settings(parser: settngs.Manager, plugins: Mapping[str, ComicTalker]) -> None:
     for talker_name, talker in plugins.items():
         try:
-            parser.add_group(talker_name, talker.register_settings, False)
+            parser.add_group("talker_" + talker_name, talker.register_settings, False)
         except Exception:
             logger.exception("Failed to register settings for %s", talker_name)
--- a/comictaggerlib/ctoptions/talker_plugins.py
+++ b/comictaggerlib/ctoptions/talker_plugins.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import logging
+from collections.abc import Mapping
 
 import settngs
 
@@ -9,9 +10,9 @@ from comictalker.talkerbase import ComicTalker
 logger = logging.getLogger(__name__)
 
 
-def register_talker_settings(parser: settngs.Manager, plugins: dict[str, ComicTalker]) -> None:
+def register_talker_settings(parser: settngs.Manager, plugins: Mapping[str, ComicTalker]) -> None:
     for talker_name, talker in plugins.items():
         try:
-            parser.add_group(talker_name, talker.comic_settings, False)
+            parser.add_group(talker_name, talker.register_settings, False)
         except Exception:
             logger.exception("Failed to register settings for %s", talker_name)
--- a/comictaggerlib/ctoptions/talker_plugins.py
+++ b/comictaggerlib/ctoptions/talker_plugins.py
@@ -1,0 +1,17 @@
+from __future__ import annotations
+
+import logging
+
+import settngs
+
+from comictalker.comictalkerapi import TalkerPlugin
+
+logger = logging.getLogger(__name__)
+
+
+def register_talker_settings(parser: settngs.Manager, plugins: dict[str, TalkerPlugin]) -> None:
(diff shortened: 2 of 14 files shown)


lordwelch:
I'm not going to review this unless the tests affected by this PR are passing. If there is a particular test that you can't get to pass give me a mention and I will look at it.

mizaki:
I've added a fix for the tests but it may be in the wrong place and the wrong way to do it too? 

It brings back the question of a new round of tests for the `talkerapi` but it is still probably better to wait until it is more finalised.

comictaggerlib/ctoptions/file.py
```diff
@@ -257,13 +239,17 @@ def validate_settings(options: settngs.Config[settngs.Values], parser: settngs.M
...
 
-def register_settings(parser: settngs.Manager) -> None:
+def register_settings(parser: settngs.Manager, talkers: dict[str, Any]) -> None:
     parser.add_group("general", general, False)
     parser.add_group("internal", internal, False)
     parser.add_group("identifier", identifier, False)
     parser.add_group("dialog", dialog, False)
     parser.add_group("filename", filename, False)
-    parser.add_group("comicvine", comicvine, False)
+    parser.add_group("talkers_general", talkers_general, False)
```
lordwelch: This can just be changed to "talkers" we don't need the extra long name

comictaggerlib/ctoptions/file.py
```diff
@@ -257,13 +239,17 @@ def validate_settings(options: settngs.Config[settngs.Values], parser: settngs.M
...
     parser.add_group("cbl", cbl, False)
     parser.add_group("rename", rename, False)
     parser.add_group("autotag", autotag, False)
+
+    # Register talker plugin settings
+    for talker, cls in talkers.items():
+        parser.add_group(talker, cls.comic_settings, False)
```
lordwelch: This needs to happen elsewhere.
This file originally only contained settings that were saved in the config file, now that it's a simple boolean that changes that it no longer strictly adheres to that but there's no reason for plugins to be passed in here.

comictaggerlib/main.py
```diff
@@ -87,7 +88,7 @@ def register_options(self) -> None:
             "For more help visit the wiki at: https://github.com/comictagger/comictagger/wiki",
         )
         ctoptions.register_commandline(self.manager)
-        ctoptions.register_settings(self.manager)
+        ctoptions.register_settings(self.manager, self.talker_plugins)
```
lordwelch: To follow the comment in file.py don't shoehorn the talker plugins into `register_settings` create a new function `register_talker_settings`

comictaggerlib/main.py
```diff
@@ -99,6 +100,18 @@ def parse_options(self, config_paths: ctoptions.ComicTaggerPaths) -> None:
         self.options = ctoptions.validate_settings(self.options, self.manager)
         self.options = self.options
 
+        # parse talker settings
+        for loaded in self.talker_plugins.values():
+            parse_options = getattr(loaded, "parse_settings", None)
```
lordwelch: what's with the naming convention switch?

lordwelch (changes requested):
If we reduce the `__init__` to:
```python
def __init__(self, version: str, cache_folder: pathlib.Path) -> None:
```

That makes it so that we can always create the object, and it should always succeed in a proper environment, so we can immediately create the object eg `talker(version, cache_folder)`.

We can even pass the `version` and `cache_folder` into `get_talkers` and then loop over all of them:
```python
def get_talkers(version, cache_folder):
  talkers = {}
  for talker in [comictalker.talkers.comicvine.ComicVineTalker]:
    try:
      object = talker(version, cache_folder)
      talkers[object.name] = object
    except Exception:
      logger.exception("Failed to load talker")
  return talkers
```
I would probably create a separate function so that people can easily get the class directly.
(though they can always use `object.__class__` for that)

Then we can make a configure (or parse_settings, etc...) method:
```python
def configure(self, config: dict[str, Any]) -> None:
```

I know flake8 used class methods but those plugins are also for the most part single function, they are given file content (as an ast or something else) and process it and spit out the answer. With the Talkers we are wrapping an entire api that we have to call multiple times to get anything useful and we may change it's state in between function calls. So I would rather keep all of the state on the instance (in `__init__`) than using class variables.

mizaki:
With https://github.com/comictagger/comictagger/pull/419/commits/0d69ba3c494a520304632f8b404357867fe4ee9d I should have addressed all the issues. I found a bit of a chicken and egg situation and decided on the approach here given your previous comments. I'll give a brief explanation:

Because of the `init` requirement of a cache path to create a talker instance the use of a class method was required to register the talker(s) options. So, to begin with all talker classes are gathered to be passed to `ctoptions.register_talker_settings` to generate the options. After the options have been parsed, `initialize_talkers` is called and `self.plugins` (in main) is propagated with the obj in addition to the class. (It may make more sense to just overwrite the class with the object and get back to a simple dict of talkers again?)

When the object is created `set_settings` (needs a name change?) is run on the talker setting the settings. If that fails, the defaults will remain. It's possible it should hard fail (and if the GUI will be used enable the users to go to settings to pick a different talker)?

The other option as I see it is the remove the cache requirement from `init` and then the objects can be created before `register_options` is called and then `initialize_talkers` can be used to set the cache along with all the other settings.

Hope all that makes sense!

mizaki:
Another thing to decide is which options should be talker settings and which should be global.

`series-match-search-thresh` - I'm leaning more global. I can see how the user may want one talker to be more forgiving than another though.

`clear-metadata`, `sort-series-by-year`, `exact-series-matches-first`, `clear-form-before-populating`, 
`use-series-start-as-volume` - All same as above.

lordwelch:
> Because of the `init` requirement of a cache path to create a talker instance the use of a class method was required to register the talker(s) options.

opts.config.user_cache_dir
here's a patch to move initialize_dirs up a 2 lines [Patch.patch](https://github.com/comictagger/comictagger/files/10502950/Patch.patch)

> When the object is created `set_settings` (needs a name change?) is run on the talker setting the settings. If that fails, the defaults will remain. It's possible it should hard fail (and if the GUI will be used enable the users to go to settings to pick a different talker)?

Hard fail, put it in the log, the developer of the talker can fix it.

> Another thing to decide is which options should be talker settings and which should be global.
> 
> `series-match-search-thresh` - I'm leaning more global. I can see how the user may want one talker to be more forgiving than another though.
> 
> `clear-metadata`, `sort-series-by-year`, `exact-series-matches-first`, `clear-form-before-populating`, `use-series-start-as-volume` - All same as above.

Currently `series-match-search-thresh` and `use-series-start-as-volume` must be implemented by the talker. The others the Talker can't control so I don't see why they would be a per-talker option. If it proves to be useful we can change it in a separate PR

mizaki:
Turns out `series-match-search-thresh` is used outside of the talker so I had to put it back in the `talkers` options. Similar to the others, as you said, if people want it, it can be done later.

Now talker instance(s) are created and `initialize_talkers` only applies the settings and will cause an exit if it fails.

comictaggerlib/main.py
```diff
@@ -66,12 +66,15 @@ def __init__(self) -> None:
...
+        self.talker_plugins: dict[str, Any] = {}
 
     def run(self) -> None:
         opts = self.initialize()
+        self.initialize_dirs(opts.config)
+        self.talker_plugins = ct_api.get_talkers(version, opts.config.user_cache_dir)
         self.register_options()
         self.parse_options(opts.config)
-        self.initialize_dirs()
+        self.initialize_talkers()
```
lordwelch: Move this to `self.parse_options` it should be called directly before `self.manager.get_namespace` and each talker should get the dictionary referring to just the settings they defined

comictalker/comictalkerapi.py
```diff
@@ -15,24 +15,36 @@
...
+    try:
+        talker.set_settings(settings)
+    except Exception as e:
+        logger.exception(
+            f"Failed to set talker settings for {talker.source_details.name}, will use defaults. Error: {str(e)}",
+        )
+        raise TalkerError(source=talker.source_details.name, code=4, desc="Could not apply talker settings")
+
+
+def get_talkers(version: str, cache: pathlib.Path) -> dict[str, Any]:
```
lordwelch: `Any` should be `ComicTalker`

comictalker/talkerbase.py
```diff
@@ -137,27 +139,20 @@ def __init__(self, source: str = "", sub_code: int = 0, desc: str = "") -> None:
...
-        self.api_key = api_key or self.default_api_key
-        self.api_url = api_url or self.default_api_url
-
-        tmp_url = urlsplit(self.api_url)
-
-        # joinurl only works properly if there is a trailing slash
-        if tmp_url.path and tmp_url.path[-1] != "/":
-            tmp_url = tmp_url._replace(path=tmp_url.path + "/")
+    def comic_settings(self, parser: settngs.Manager) -> None:
+        """Talker settings."""
```
lordwelch: This should be `register_settings`.
docstring should be along the lines of

> Allows registering settings using the settngs package using a argparse like interface

comictalker/talkerbase.py
```diff
@@ -137,27 +139,20 @@ def __init__(self, source: str = "", sub_code: int = 0, desc: str = "") -> None:
...
 
-        self.api_url = tmp_url.geturl()
+    def set_settings(self, settings: argparse.Namespace) -> None:
+        """Apply talker settings from config to object."""
```
lordwelch: This should be `def parse_settings(self, settings: dict[str, Any])`

>  `settings` is a dictionary of options defined in `register_settings`.
> It is only guaranteed that all settings defined in `register_settings` will be present, there may be other options present in the dictionary.

comictalker/talkers/comicvine.py
```diff
@@ -176,19 +169,66 @@ def __init__(
...
+        parser.add_setting(
+            "--cv-url",
+            help="Use the given Comic Vine URL.",
+        )
 
-        self.series_match_thresh: int = series_match_thresh
+    def set_settings(self, settings: argparse.Namespace) -> None:
+        """Set settings."""
+        if settings.comicvine_cv_remove_html_tables:
+            self.remove_html_tables = bool(settings.comicvine_cv_remove_html_tables)
```
lordwelch: https://www.youtube.com/watch?v=TJ_aGY6iyZY

comictalker/talkers/comicvine.py
```diff
@@ -176,19 +169,66 @@ def __init__(
...
+        if settings.comicvine_cv_url:
+            try:
+                tmp_url = urlsplit(settings.comicvine_cv_url)
+                # joinurl only works properly if there is a trailing slash
+                if tmp_url.path and tmp_url.path[-1] != "/":
+                    tmp_url = tmp_url._replace(path=tmp_url.path + "/")
+
+                self.api_url = tmp_url.geturl()
+            except Exception:
+                logger.exception("Failed to parse new talker URL for %s, will use default", self.source_name)
```
lordwelch: This should not be caught. If there is an error it should be handled in ComicTagger.

comictalker/comictalkerapi.py
mizaki: I had to change this to Mapping because mypy and `"Dict" is invariant`. Is there a better way?

comictalker/talkers/comicvine.py
mizaki: As any bool setting should have already been checked, I changed it to just equalling the value.

comictalker/comictalkerapi.py
lordwelch: The `talkers` variable on L41 needed to be typed as well
`talkers: dict[str, ComicTalker] = {}`

comictalker/comictalkerapi.py
```diff
@@ -16,23 +16,36 @@
...
-    """Returns all comic talker modules NOT objects"""
-    return {"comicvine": comictalker.talkers.comicvine.ComicVineTalker}
+def set_talker_settings(talker: ComicTalker, settings: dict[str, Any]) -> None:
+    try:
+        talker.parse_settings(settings)
+    except Exception as e:
+        logger.exception(
+            f"Failed to set talker settings for {talker.source_details.name}, will use defaults. Error: {str(e)}",
+        )
+        raise TalkerError(source=talker.source_details.name, code=4, desc="Could not apply talker settings")
```
lordwelch: What's the point of this function?
Currently the only place this is used does this
```python
try:
    for talker_name, talker in self.talker_plugins.items():
        ct_api.set_talker_settings(talker, self.options[0][talker_name])
except Exception as e:
    # Remove talker as we failed to apply the settings
    del self.talker_plugins[e.source]  # type: ignore[attr-defined]
    logger.exception("Failed to initialize talker settings. Error %s", str(e))
```
Which does not benefit from anything that this function does and catches the wrong exceptions and only removes the first talker that fails.

It should be:
```python
for talker_name, talker in list(self.talker_plugins.items()):
    try:
        ct_api.set_talker_settings(talker, self.options[0][talker_name])
    except ct_api.TalkerError as e:
        # Remove talker as we failed to apply the settings
        del self.talker_plugins[e.source]
        logger.exception("Failed to initialize talker settings. Error %s", e)
```

Which can entirely be replaced with this ignoring the `set_talker_settings` function:
```python
for talker_name, talker in list(self.talker_plugins.items()):
    try:
        talker.parse_settings(self.options[0][talker_name])
    except Exception as e:
        # Remove talker as we failed to apply the settings
        del self.talker_plugins[talker_name]
        logger.exception("Failed to initialize talker settings. Error %s", e)
```

comictaggerlib/ctoptions/talker_plugins.py
```diff
@@ -0,0 +1,18 @@
...
+
+from comictalker.talkerbase import ComicTalker
+
+logger = logging.getLogger(__name__)
+
+
+def register_talker_settings(parser: settngs.Manager, plugins: Mapping[str, ComicTalker]) -> None:
+    for talker_name, talker in plugins.items():
+        try:
+            parser.add_group(talker_name, talker.register_settings, False)
```
lordwelch: I want all of these groups to have `talker_` as a prefix in the group name. I don't want it to be possible to have a name conflict with a ComicTagger group. If two talkers have a conflict they can deal with it themselves.

comictaggerlib/main.py
```diff
@@ -136,32 +151,14 @@ def main(self) -> None:
...
-        # None comparison is used so that the empty string can unset the value
-        if self.options[0].comicvine_cv_api_key is not None or self.options[0].comicvine_cv_url is not None:
-            settings_path = self.options[0].runtime_config.user_config_dir / "settings.json"
-            if self.config_load_success:
-                self.manager.save_file(self.options[0], settings_path)
-
-        if self.options[0].commands_only_set_cv_key:
-            if self.config_load_success:
-                print("Key set")  # noqa: T201
-                return
```
lordwelch: This should not be removed

comictaggerlib/main.py
```diff
@@ -136,32 +151,14 @@ def main(self) -> None:
...
-                wait_on_ratelimit=self.options[0].autotag_wait_and_retry_on_rate_limit,
-                api_url=self.options[0].comicvine_cv_url,
-                api_key=self.options[0].comicvine_cv_api_key,
-            )
-        except TalkerError as e:
-            logger.exception("Unable to load talker")
-            error = (str(e), True)
+            talker_api = self.talker_plugins[self.options[0].talkers_source]
+        except Exception as e:
+            logger.exception(f"Unable to load talker {self.options[0].talkers_source}. Error: {str(e)}")
```
lordwelch: remove the `str(e)`

comictaggerlib/main.py
```diff
@@ -136,32 +151,14 @@ def main(self) -> None:
...
+            # TODO error True can be changed to False after the talker settings menu generation is in
+            error = (f"Unable to load talker {self.options[0].talkers_source}. Error: {str(e)}", True)
```
lordwelch: remove the `str(e)`

comictalker/talkerbase.py
```diff
@@ -137,27 +138,21 @@ def __init__(self, source: str = "", sub_code: int = 0, desc: str = "") -> None:
...
-        self.api_key = api_key or self.default_api_key
-        self.api_url = api_url or self.default_api_url
-
-        tmp_url = urlsplit(self.api_url)
-
-        # joinurl only works properly if there is a trailing slash
-        if tmp_url.path and tmp_url.path[-1] != "/":
-            tmp_url = tmp_url._replace(path=tmp_url.path + "/")
+    def register_settings(self, parser: settngs.Manager) -> None:
+        """Allows registering settings using the settngs package with an argparse like interface"""
```
lordwelch, line 153: Add a `return None`

comictalker/talkerbase.py
```diff
@@ -137,27 +138,21 @@ def __init__(self, source: str = "", sub_code: int = 0, desc: str = "") -> None:
...
 
-        self.api_url = tmp_url.geturl()
+    def parse_settings(self, settings: dict[str, Any]) -> None:
+        """settings is a dictionary of options defined in register_settings.
+        It is only guaranteed that the settings defined in register_settings will be present."""
```
lordwelch, line 158: Add a `return None`

lordwelch (changes requested):

comictalker/comictalkerapi.py
mizaki: I think it's where it was leftover and I didn't think it through.

PR #420, opened 2023-01-05
Comic Source Attribution

mizaki:
This is the basic version of recognising the source of the comic data. It's possible that this would also be desired/required (Marvel) for the issues window too.

Right now the URL doesn't open a browser window. I believe this would require import `webbrowser` and some minimal code ala `taggerwindow` etc. It may be better to add the URL checking code to `utils`(?) to eliminate any duplication of code?

I've gone with a local file as that seemed safest. It may be better to be file/URL agnostic though?


mizaki:
Screenshot:

Screenshot_20230106_213553

comictaggerlib/seriesselectionwindow.py
```diff
@@ -161,6 +161,17 @@ def __init__(
         # Load to retrieve settings
         self.talker_api = talker_api
 
+        # Display talker logo and set url
+        self.lblSourceName.setText(
+            f'<html><head/><body><p>Data Source: <a href="{talker_api.static_options.website}"><span style=" text-decoration: underline; color:#5d88ca;">{talker_api.static_options.website}</span></a></p></body></html>'
+        )
```
lordwelch: The only html you need here is the href

comictaggerlib/ui/seriesselectionwindow.ui
```diff
@@ -85,17 +146,11 @@
...
-            <set>AlignCenter</set>
-           </property>
           </column>
           <column>
            <property name="text">
             <string>Issues</string>
            </property>
-           <property name="textAlignment">
-            <set>AlignCenter</set>
-           </property>
```
lordwelch, line 98: Is there a reason text alignment is changing here?

comictaggerlib/seriesselectionwindow.py
```diff
@@ -161,6 +161,17 @@ def __init__(
...
+        source_label_logo = QtGui.QPixmap(talker_api.source_details.logo)
+        if source_label_logo.height() > 100 or source_label_logo.width() > 300:
+            source_label_logo = source_label_logo.scaled(
+                300, 100, QtCore.Qt.KeepAspectRatio, QtCore.Qt.SmoothTransformation
+            )
+        self.lblSourceLogo.setPixmap(source_label_logo)
```
lordwelch: Instead of doing all of this, reuse the coverimage widget it takes care of all of this.

lordwelch (changes requested):
As this will only be shown when searching, which by definition must be online, lets use a url it will make for smaller changes. If the CoverImageWidget is used it will even cache the image.

comictaggerlib/ui/seriesselectionwindow.ui
mizaki: I believe this is because `center` is the default and so Qt Designer removed it when I saved as it's superfluous. If I open it from `develop` and save, it does the same thing.

lordwelch: I must have an old version of QT Designer...

mizaki:
There seems to be some margin space thing happening at the bottom of the image with `imageWidget`? If you resize the window downwards, you should see. Is that `imageWidget` or something else do you know?

Also, I check the Marvel terms again:
```
You must attribute Marvel as the source of data whenever you display any results from the Marvel Comics API. Please use the following text on every application screen or web page which displays the API result:
"Data provided by Marvel. © 2014 Marvel"
```

To me that reads as the issues window also needs the same (minus the logo?). Add a `copyright` str to the talkers `source_details` too for the data source text? I imagine Marvel would be the most demanding so if they are covered, every other source would be too.

lordwelch:
It's probably `CoverImageWidget` unless it obscures something or makes it unusable we can open an issue when this is merged and find a fix for it later.

Yeah, go make the same changes on the Issue window.
Let's call the attribute `attribution` or `attribution_string` , line, etc

comictalker/talkerbase.py
```diff
@@ -40,13 +40,15 @@ class SourceStaticOptions:
     def __init__(
         self,
         website: str = "",
+        attribution_string: str = "",  # Website link will be added after this text using source_details.name
```
lordwelch: Let's make the attribution_string stand by itself. This string should always make sense by itself. Maybe put a comment that the html anchor tag is understood.

comictaggerlib/ui/seriesselectionwindow.ui
```diff
@@ -17,23 +17,93 @@
...
+         <property name="midLineWidth">
+          <number>1</number>
+         </property>
+         <property name="orientation">
+          <enum>Qt::Horizontal</enum>
+         </property>
+        </widget>
+       </item>
+       <item>
+        <widget class="QLabel" name="lblSourceName">
```
lordwelch, line 69: Check the openExternalLinks. box so that the links will work

lordwelch (changes requested):

comictalker/talkerbase.py
```diff
@@ -40,13 +40,15 @@ class SourceStaticOptions:
     def __init__(
         self,
         website: str = "",
+        attribution_string: str = "",  # Full string including web link: Source: <a href='http://website'>Example</a>
```
lordwelch: Let's use a more verbose default:
> Metadata provided by &lt;a href='https[]()://comicvine.gamespot.com/'>Comic Vine</a>.

lordwelch (approved):

PR #425, opened 2023-02-07
Hide coverimagewidget text when there is a singles image. Fixes #424

### LINKED ISSUE #424: CoverImageWidget always leave a space at the bottom
It appears that CoverImageWidget will always leave space for the `next/previous` controls at the bottom of the widget even when there is only a single image.

mizaki:
It might be intended that `showControls = False` is meant to be used instead? The question is with say the issue view, is it preferred that the image scales to the size of the widget or leave the space for the controls?

If using `showControls` would it not be better to have it passed in on `init`?

### DIFF
--- a/comictaggerlib/coverimagewidget.py
+++ b/comictaggerlib/coverimagewidget.py
@@ -218,11 +218,13 @@ class CoverImageWidget(QtWidgets.QWidget):
             self.btnRight.setEnabled(False)
             self.btnLeft.hide()
             self.btnRight.hide()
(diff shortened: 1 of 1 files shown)


mizaki:
Here are some screenshots to help illustrate.
Screenshot_20230208_164739
No space left for controls.

Screenshot_20230208_164719
Space for controls but hidden

Screenshot_20230208_164646
Controls

The other option is change it to floating arrows:
Screenshot_20230208_164739_float

PR #430, opened 2023-02-11
Generate settings tabs for each talker

mizaki:
Mostly based on the previous work. Changed passing in the current talker to the whole list for the settings window.


comictaggerlib/settingswindow.py
```diff
@@ -235,6 +234,100 @@ def disconnect_signals(self) -> None:
         self.twLiteralReplacements.cellChanged.disconnect()
         self.twValueReplacements.cellChanged.disconnect()
 
+        self.sources: dict = {}
+        self.generate_source_option_tabs()
+
+    def generate_source_option_tabs(self) -> None:
```
lordwelch: I want this function in it's own file and pass in as an argument the widget that everything will be built into, the config, and the talkers

comictaggerlib/settingswindow.py
```diff
@@ -235,6 +234,100 @@ def disconnect_signals(self) -> None:
...
+        def format_internal_name(int_name: str = "") -> str:
```
lordwelch: I want a test written for this so I can see exactly the end result should look like

comictaggerlib/settingswindow.py
```diff
@@ -235,6 +234,100 @@ def disconnect_signals(self) -> None:
...
+            self.cobxInfoSource.addItem(talker_obj.name, talker_id)
+
+            # Use a dict to make a var name from var
+            source_info = {}
+            tab_name = talker_id
+            source_info[tab_name] = {"tab": QtWidgets.QWidget(), "widgets": {}}
+            layout_grid = QtWidgets.QGridLayout()
+            row = 0
+
+            full_talker_name = "talker_" + talker_id
```
lordwelch: This variable name is longer than the content and `full_talker_name` sounds like it's the display name.
Using `f"talker_{talker_id}"` Mae's it much easier to see what it is

comictaggerlib/settingswindow.py
```diff
@@ -235,6 +234,100 @@ def disconnect_signals(self) -> None:
...
+            for option in self.config[1][full_talker_name][1].values():
+                current_widget = None
+                if option.action is not None and isinstance(option.action, type(argparse.BooleanOptionalAction)):
```
lordwelch: I want each of these to be split into separate functions. Also add store_true and store_false from argparse to this

comictaggerlib/settingswindow.py
```diff
@@ -235,6 +234,100 @@ def disconnect_signals(self) -> None:
...
+                row += 1
+
+                if current_widget:
+                    # Add tooltip text
+                    current_widget.setToolTip(option.help)
+
+                    source_info[tab_name]["widgets"][option.internal_name] = current_widget
+                else:
+                    # An empty current_widget implies an unsupported type
+                    logger.info(f"Unsupported talker option found. Name: {option.internal_name} Type: {option.type}")
```
lordwelch: Make this a debug log

lordwelch (changes requested):
I think I also want to do some extra processing for the API key but I'm not sure what.

mizaki:
Possibly a lot of issues:
- File location. Seemed to make sense as it was UI related but...?
- File name?
- Should the menu generator be a class?
- I feel like there should be an easier way with the test api/url functions (making the whole thing a class?).
- Adding separate functions for checking API and URL made sense to me.
- Not sure on the benefit of moving the specific widget generator function out on their own? (Probably missing something.)

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,213 @@
...
+    parent: QtWidgets.QWidget,
+):
+    key = ""
+    # Find the correct widget to get the API key
+    for name, widget in sources_info[talker_id]["widgets"].items():
+        if name.startswith("talker_" + talker_id) and name.endswith("api_key"):
+            key = widget.text().strip()
+
+    if talkers[talker_id].check_api_key(key):
+        QtWidgets.QMessageBox.information(parent, "API Key Test", "Key is valid!")
```
lordwelch: Parent is not needed `None` works for all of the QMessageBox calls.
Even if `None` didn't work, it doesn't matter which QWidget is passed in so we don't need a separate QWidget piped into this we can just use the button that triggers this function

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,213 @@
...
+        if name.startswith("talker_" + talker_id) and name.endswith("url"):
+            url = widget.text().strip()
+
+    if talkers[talker_id].check_api_url(url):
+        QtWidgets.QMessageBox.information(parent, "API Key Test", "URL is valid!")
+    else:
+        QtWidgets.QMessageBox.warning(parent, "API Key Test", "URL is NOT valid!")
+
+
+def test_api_key(
```
lordwelch: This should not be `test_api_key` this does no testing it only connects a function to a button. Same with `test_api_url`

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,213 @@
...
+def format_internal_name(int_name: str = "") -> str:
+    # Presume talker_<name>_<nm>: talker_comicvine_cv_widget_name
+    int_name_split = int_name.split("_")
+    del int_name_split[0:3]
+    int_name_split[0] = int_name_split[0].capitalize()
+    new_name = " ".join(int_name_split)
+    return new_name
+
+
+def generate_checkbox(
```
lordwelch: For generating these widgets they should not have a value to start with, they should start unconfigured and and have a separate function to set the values so that it goes in line with the rest of the settings window

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,213 @@
...
+        # Use a dict to make a var name from var
+        source_info = {}
+        tab_name = talker_id
+        source_info[tab_name] = {"tab": QtWidgets.QWidget(), "widgets": {}}
+        layout_grid = QtWidgets.QGridLayout()
+
+        for option in config[1][f"talker_{talker_id}"][1].values():
+            current_widget = None
+            if option.action is not None and (
+                isinstance(option.action, type(argparse.BooleanOptionalAction))
```
lordwelch: All of these `isinstance` can be changed to `is` eg `option.action is argparse.BooleanOptionalAction`
And you should add `option.type is bool` to this one for completion

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,213 @@
...
+    """
+    Generate GUI tabs and settings for talkers
+    """
+
+    sources: dict = {}
+
+    # Add source sub tabs to Comic Sources tab
+    for talker_id, talker_obj in talkers.items():
+        # Add source to general tab dropdown list
+        tabs.findChildren(QtWidgets.QComboBox, "cobxInfoSource")[0].addItem(talker_obj.name, talker_id)
```
lordwelch: This findChildren should definitely not be in this loop at a minimum.
but should actually be generated along with the rest of these options.
```suggestion
    cobxInfoSource = tabs.findChildren(QtWidgets.QComboBox, "cobxInfoSource")[0]
    # Add source sub tabs to Comic Sources tab
    for talker_id, talker_obj in talkers.items():
        # Add source to general tab dropdown list
        cobxInfoSource.addItem(talker_obj.name, talker_id)
```

comictaggerlib/ui/settingswindow.ui
```diff
@@ -391,100 +327,127 @@
...
-              <property name="text">
-               <string>Comic Vine URL</string>
-              </property>
-             </widget>
-            </item>
-            <item row="2" column="1">
-             <widget class="QLineEdit" name="leURL"/>
-            </item>
-           </layout>
+          <widget class="QTabWidget" name="tTalkerTabs">
```
lordwelch: Move all of these to the search tab. None of these options are directly about a source they are mostly about displaying search results. This allows us to have all of the content in this tab be generated in code.

comictaggerlib/settingswindow.py
```diff
@@ -189,12 +190,20 @@ def __init__(
         self.settings_to_form()
         self.rename_test()
         self.dir_test()
+        self.sources: dict = {}
+        self.sources = comictaggerlib.ui.talkeruigenerator.generate_source_option_tabs(
+            parent, self.tTalkerTabs, self.config, self.talkers
+        )
+        # Select active source in dropdown
+        self.cobxInfoSource.setCurrentIndex(self.cobxInfoSource.findData(self.config[0].talker_source))
```
lordwelch: This should happen in `settings_to_form` with everything else

comictaggerlib/settingswindow.py
```diff
@@ -189,12 +190,20 @@ def __init__(
         self.settings_to_form()
         self.rename_test()
         self.dir_test()
+        self.sources: dict = {}
+        self.sources = comictaggerlib.ui.talkeruigenerator.generate_source_option_tabs(
+            parent, self.tTalkerTabs, self.config, self.talkers
+        )
```
lordwelch:
```suggestion
        self.sources: dict = comictaggerlib.ui.talkeruigenerator.generate_source_option_tabs(
            parent, self.tTalkerTabs, self.config, self.talkers
        )
```

lordwelch (changes requested):

mizaki:
That should have addressed everything. General for talkers is looking a little bare now it only have the dropdown for the talker selection.

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+    lbl_info = QtWidgets.QLabel("Information Source:")
+    cbx_info = QtWidgets.QComboBox()
+    general_layout.addWidget(lbl_info, 0, 0, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Maximum)
+    general_layout.addWidget(cbx_info, 0, 1, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Maximum)
+    vspacer = QtWidgets.QSpacerItem(20, 40, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
+    general_layout.addItem(vspacer, 1, 0)
+    source_info["general"]["widgets"]["talker_source"] = cbx_info
+
+    source_info["general"]["tab"].setLayout(general_layout)
+    tabs.addTab(source_info["general"]["tab"], "General")
```
lordwelch: A general tab is not needed as this is now only a selector for the active Talker.
Remove tTalkerTabs from the ui file, pass in `tComicTalkers` instead.
Here create the label and ComboBox and then below it create the QTabWidget with a tab for each talker

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+
+    # Add source sub tabs to Comic Sources tab
+    for talker_id, talker_obj in talkers.items():
+        # Add source to general tab dropdown list
+        source_info["general"]["widgets"]["talker_source"].addItem(talker_obj.name, talker_id)
```
lordwelch: This isn't needed just use `cbx_info` to set and retrieve the value directly, this is not a generated value there is no reason to "find" it as it is currently hardcoded in this PR.

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+                or option.action == "store_false"
+            ):
+                layout_grid, current_widget = generate_checkbox(option, layout_grid)
+                source_info[tab_name]["widgets"][option.internal_name] = current_widget
+            elif option.type is int or option.type is float:
+                layout_grid, current_widget = generate_spinbox(
+                    option, getattr(config[0], option.internal_name), layout_grid
+                )
+                source_info[tab_name]["widgets"][option.internal_name] = current_widget
+            # option.type of None should be string
```
lordwelch: Separate this into two if statements

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+    tabs: QtWidgets.QTabWidget,
+    config: settngs.Config[settngs.Namespace],
+    talkers: dict[str, ComicTalker],
+) -> dict[str, QtWidgets.QWidget]:
+    """
+    Generate GUI tabs and settings for talkers
+    """
+
+    sources: dict = {}
+    # Use a dict to make a var name from var
```
lordwelch: This comment does not make sense

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+
+        # Add vertical spacer
+        vspacer = QtWidgets.QSpacerItem(20, 40, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
+        layout_grid.addItem(vspacer, layout_grid.rowCount() + 1, 0)
+        # Display the new widgets
+        source_info[tab_name]["tab"].setLayout(layout_grid)
+
+        # Add new sub tab to Comic Source tab
+        tabs.addTab(source_info[tab_name]["tab"], talker_obj.name)
+        sources.update(source_info)
```
lordwelch: You are doing something wrong if you are using update here.
update is for when you get a dictionary from somewhere else and need to add all of it's content to a dictionary you got from a different place. All of these dictionaries are created here in this function.
```pycon
>>> a = {'a': True}
>>> b = {'a': False, 'b': True}
>>> a.update(b)
>>> a
{'a': False, 'b': True}
>>> a = {'a': True}
>>> a['b'] = True
>>> a
{'a': True, 'b': True}
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+    # Special case for api_key, make a test button
+    if option.internal_name.endswith("api_key"):
+        btn = QtWidgets.QPushButton("Test Key")
+        layout.addWidget(btn, layout.rowCount() - 1, 2)
+
+    if option.internal_name.endswith("url"):
+        btn = QtWidgets.QPushButton("Test URL")
+        layout.addWidget(btn, layout.rowCount() - 1, 2)
+
+    return layout, widget, btn
```
lordwelch: Why is the layout returned?

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+def generate_textbox(
+    option: settngs.Setting, layout: QtWidgets.QGridLayout
+) -> tuple[QtWidgets.QGridLayout, QtWidgets.QLineEdit, QtWidgets.QPushButton]:
+    btn = None
+    # str equals a text field (QLineEdit)
+    lbl = QtWidgets.QLabel(format_internal_name(option.internal_name))
+    # Create a label
+    layout.addWidget(lbl, layout.rowCount() + 1, 0)
```
lordwelch: Rows are zero indexed, this should be replaced with this:
```python
row = layout.rowCount()
layout.addWidget(lbl, row, 0)
```

and row should be used whenever the row number is needed

tests/talker_test.py
```diff
@@ -0,0 +1,7 @@
+from __future__ import annotations
+
+import comictaggerlib.ui.talkeruigenerator
+
+
+def test_format_internal_name():
+    assert comictaggerlib.ui.talkeruigenerator.format_internal_name("talker_comicvine_cv_test_name") == "Test name"
```
lordwelch: parametrize this. See https://docs.pytest.org/en/6.2.x/parametrize.html

add "talker_comic_vine_cv2_test_name" as another test case (because someone is going to put an underscore in their talker id and there is no reason to try to stop them)

comictaggerlib/settingswindow.py
```diff
@@ -464,6 +456,22 @@ def accept(self) -> None:
...
+                    widget_value = widget.value()
+                elif isinstance(widget, QtWidgets.QLineEdit):
+                    widget_value = widget.text().strip()
+                elif isinstance(widget, QtWidgets.QCheckBox):
+                    widget_value = widget.isChecked()
+                # The talker source dropdown
+                elif isinstance(widget, QtWidgets.QComboBox):
+                    widget_value = widget.itemData(widget.currentIndex())
+
+                setattr(self.config[0], name, widget_value)
```
lordwelch: This should be a function in talkeruigenerator.py

lordwelch (changes requested):

comictaggerlib/ui/talkeruigenerator.py
mizaki: It appears that `source_info` duplicates the end result of `sources`. `source_info` was inside the tab loop so it might have made sense then. I've removed `source_info`.

comictaggerlib/ui/talkeruigenerator.py
mizaki: Brain fart. I was thinking the `layout` var would be local to the function so it would need to be passed back.

tests/talker_test.py
mizaki: I've switch to using the `dest` value from the config; `cv_test_name`, that does still assume the first block of text is a single string. I'm not sure of another way to format those strings without making that assumption. The other option could be to add something like a `display` option to `settngs` but I'm not sure it would make sense for anyone else?

It may also not be a good idea to expect `dest`?

lordwelch: using `cv_test_name` or `cv test name` would be acceptable as that is what the commandline option will be `--cv-test-name` and `cv_test_name` is what is directly used in the config file. Adding a display name might be a good idea as I can make it default to `dest` if it is not set. I might make that change later.
Until I make that change I want at least 3 test cases on this function, even if it seems trivial and all that this function does is `name.replace("_", " ").title()`

`dest` is always set and it is what is used internally to identify the options when it is a dictionary, so it is a good option to use it here.

mizaki: I've added another test case with the same assumption. If you do add a `display` option to `settngs` then obviously the function and test isn't needed any more.

comictaggerlib/ui/talkeruigenerator.py
mizaki: I've done the `cbx_info` change and also added it to the sources widget list. As for the finddata, afaik this is the best way to set the current item in the combo box? Even if there is only one talker now, there won't be so I think it makes sense to put this in now rather than say, with a second talker addition.

tests/talker_test.py
mizaki: I see you've added `display_name` so I've switched to using that.

comictaggerlib/ui/talkeruigenerator.py
lordwelch: By find I just meant iterating through the data in order to find the single combobox.

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+    if option.internal_name.endswith("url"):
+        btn = QtWidgets.QPushButton("Test URL")
+        layout.addWidget(btn, row, 2)
+
+    return widget, btn
+
+
+def settings_to_talker_form(sources: dict[str, QtWidgets.QWidget], config: settngs.Config[settngs.Namespace]) -> None:
+    # Set the active talker in sources combo box
+    sources["talker_source"].setCurrentIndex(sources["talker_source"].findData(config[0].talker_source))
```
lordwelch: This can be a single call to `setCurrentText`.
> The setter [setCurrentText]() simply calls [setEditText]\() if the combo box is editable. Otherwise, if there is a matching text in the list, [currentIndex] is set to the corresponding index.

[setCurrentText]: https://doc.qt.io/qt-5/qcombobox.html#currentText-prop
[setEditText]: https://doc.qt.io/qt-5/qcombobox.html#setEditText
[currentIndex]: https://doc.qt.io/qt-5/qcombobox.html#currentIndex-prop

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+                    widget.setValue(value)
+                if value_type is bool:
+                    widget.setChecked(value)
+            except Exception:
+                logger.debug("Failed to set value of %s", name)
+
+
+def form_settings_to_config(sources: dict[str, QtWidgets.QWidget], config: settngs.Config[settngs.Namespace]) -> None:
+    # Source combo box value
+    config[0].talker_source = sources["talker_source"].itemData(sources["talker_source"].currentIndex())
```
lordwelch: This can be a single call to `currentText`.

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+
+        tab_name = talker_id
+        sources["tabs"][tab_name] = {"tab": QtWidgets.QWidget(), "widgets": {}}
+        layout_grid = QtWidgets.QGridLayout()
+
+        for option in config[1][f"talker_{talker_id}"][1].values():
+            current_widget = None
+            if option.action is not None and (
+                option.action is argparse.BooleanOptionalAction
+                or option.action is bool
```
lordwelch:
```suggestion
                or option.type is bool
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+                current_widget = generate_checkbox(option, layout_grid)
+                sources["tabs"][tab_name]["widgets"][option.internal_name] = current_widget
+            elif option.type is int:
+                current_widget = generate_spinbox(option, layout_grid)
+                sources["tabs"][tab_name]["widgets"][option.internal_name] = current_widget
+            elif option.type is float:
+                current_widget = generate_doublespinbox(option, layout_grid)
+                sources["tabs"][tab_name]["widgets"][option.internal_name] = current_widget
+            # option.type of None should be string
+            elif option.type is None or option.type is str:
```
lordwelch:
```suggestion
            elif (option.type is None and option.action is None) or option.type is str:
```

lordwelch (changes requested):
The last thing is I really don't like how the api key and URL are setup, it doesn't make sense to test the url separate from the API key, I know it makes generating the UI easier but I still don't like it. I'm planning on pushing a commit today to make it consistent across all talkers.

comictaggerlib/ui/talkeruigenerator.py
mizaki: That will return the name (Comic Vine) and not the talker_id (comicvine) which is in the item data and is what needs to be saved to config.

comictaggerlib/ui/talkeruigenerator.py
mizaki: My preference for the findData is that is will be unique (talker_id). Also will need to pass in the name (Comic Vine) from talkers (not a hardship but still...). Granted there shouldn't be two talkers with the same title but this is more robust. I can change it if you really want though?

mizaki:
> The last thing is I really don't like how the api key and URL are setup, it doesn't make sense to test the url separate from the API key, I know it makes generating the UI easier but I still don't like it. I'm planning on pushing a commit today to make it consistent across all talkers.

It's not so much about making it easier to generate, more that most APIs don't require a key so there will only be a URL to test. On that note, it looks like you've made every talker have an API key entry?

comictaggerlib/ui/talkeruigenerator.py
mizaki: Looks like there is currentData() is that what you meant? Seems to be :)

comictaggerlib/ui/talkeruigenerator.py
lordwelch: I thought this was the ID not the name.
Also lets change "talker_source" to "cbxTalker" or "TalkerComboBox" or something similar

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+    for talker_id, talker_obj in talkers.items():
+        # Add source to general tab dropdown list
+        cbx_info.addItem(talker_obj.name, talker_id)
+
+        tab_name = talker_id
+        sources["tabs"][tab_name] = {"tab": QtWidgets.QWidget(), "widgets": {}}
+        layout_grid = QtWidgets.QGridLayout()
+
+        for option in config[1][f"talker_{talker_id}"][1].values():
+            current_widget = None
```
lordwelch: > It's not so much about making it easier to generate, more that most APIs don't require a key so there will only be a URL to test. On that note, it looks like you've made every talker have an API key entry?

I haven't checked all of the different comic/manga websites for what their API requires, but in general most websites require that you register an account to make API queries.

Now that the key and url settings always exist, an arbitrary extension check (`.endswith("key")`) is not needed.
All of them are `{id}_key` and `{id}_url` so this is known before hand and can be added with only one button, how it was before. Before looping over the settings definition `config[1]` the key and url text boxes can be put in with a single button to test the API and then skip those two names when generating the rest of the settings.
If they don't wan't the settings to show up at all they can use
```python
        parser.add_setting(f"--{self.id}-key", file=False, cmdline=False)
        parser.add_setting(f"--{self.id}-url", file=False, cmdline=False)
```
and it will not show up in any settings (if there is already one saved in the settings file the value will stay but will be unchangeable from the CLI or the settings window) , the only place it will show up is on the definitions

And this check needs to be added, if a talker doesn't plan on saving the setting then it doesn't need to be in the GUI.

```suggestion
        for option in config[1][f"talker_{talker_id}"][1].values():
            if not option.file:
                continue
            current_widget = None
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+
+    # Store all widgets as to allow easier access to their values vs. using findChildren etc. on the tab widget
+    sources: dict = {"tabs": {}}
+
+    # Tab comes with a QVBoxLayout
+    comic_talker_tab_layout = comic_talker_tab.layout()
+
+    talker_layout = QtWidgets.QGridLayout()
+    lbl_info = QtWidgets.QLabel("Information Source:")
+    cbx_info = QtWidgets.QComboBox()
```
lordwelch: This needs to be "Metadata Source:"
and lbl_info and cbx_info should be *_talker and should keep the same name in the sources dict

comictaggerlib/ui/settingswindow.ui
```diff
@@ -308,197 +336,11 @@
          </item>
         </layout>
        </widget>
-       <widget class="QWidget" name="tComicVine">
+       <widget class="QWidget" name="tComicTalkers">
         <attribute name="title">
-         <string>Comic Vine</string>
+         <string>Comic Sources</string>
```
lordwelch: Lets change this to `Metadata Sources` just to make it clear what is being retrieved.

comictaggerlib/ui/talkeruigenerator.py
mizaki: Just on the API key thing: Manga Updates, Jikan, kitsu do not require an API key for getting data. For other things they does but we don't care about those. I presume they all limit by IP if someone gets too greedy.

mizaki: An additional small gripe is that the URL and key are always on top in the GUI now. Before the order was dictated by the order within the talker. It's a personal preference but I think it looks better when the check boxes are first. I could order them with the generation but that takes the order out of the talkers hands as well. Very minor but seeing a text field followed by check boxes and then possibly another text field will hurt a little :)

Now you've added a `display_name` to settngs, is an `order` field worth any consideration?

mizaki: Another reason for separate test buttons and functions is not knowing which one has failed. More logic could be placed in the test on the talker and return a message rather than True/False.

If going with only one button, it would have to go on the URL line (because key is optional with some talkers). It would then make sense to have a tool tip explaining it will test both.

Does all that check out?

lordwelch: I don't have an issue either way they just need to be next to each other you can easily just add it after you add everything else.

No, not adding an order field, plugins can manage order by what order they call `add_setting`.

That's something that can be worked on later, currently that's all it gives for ComicVine, it will work for any others for now, and they can always write to the log.

I'm sure someone will decide the url is optional but not the key.
```python
def add_api_widgets(config: settngs.Config[settngs.Namespace], layout: QtWidgets.QGridLayout, talker: ComicTalker):
    # *args enforces keyword arguments and allows position arguments to be ignored
    def call_check_api(*args, le_url: QtWidgets.QLineEdit, le_key: QtWidgets.QLineEdit, talker: ComicTalker):
        url = ""
        key = ""
        if le_key is not None:
            key = le_key.text().strip()
        if le_url is not None:
            url = le_url.text().strip()
        if talker.check_api_key(key, url):
            QtWidgets.QMessageBox.information(None, "API Test", "Key is valid!")
        else:
            QtWidgets.QMessageBox.warning(None, "API Test", "Key is NOT valid!")

    # get the actual config objects in case they have overwritten the default
    talker_key = config[1][f"talker_{talker.id}"][1][f"{talker.id}_key"]
    talker_url = config[1][f"talker_{talker.id}"][1][f"{talker.id}_url"]
    btn_test_row = None
    le_key = None
    le_url = None

    # only file settings are saved
    if talker_key.file:
        # record the current row so we know where to add the button
        btn_test_row = layout.rowCount()
        le_key = generate_textbox(talker_key, layout)

    # only file settings are saved
    if talker_url.file:
        # record the current row so we know where to add the button
        # We overwrite so that the default will be next to the url text box
        btn_test_row = layout.rowCount()
        le_url = generate_textbox(talker_url, layout)

    # The button row was recorded so we add it
    if btn_test_row is not None:
        btn = QtWidgets.QPushButton("Test API")
        layout.addWidget(btn, btn_test_row, 2)
        # partial is used as connect will pass in event information
        btn.clicked.connect(partial(call_check_api, le_url=le_url, le_key=le_key, talker=talker))
```

Then you call `add_api_widgets(config, layout_grid, talker_obj)` either before or after the rest of the widgets and you add this to the beginning of the widget generation loop:
```python
            if option.dest in (f"{talker_id}_url", f"{talker_id}_key"):
                continue
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,239 @@
...
+    widget = QtWidgets.QCheckBox(option.display_name)
+    widget.setToolTip(option.help)
+    layout.addWidget(widget, layout.rowCount(), 0, 1, -1)
+
+    return widget
+
+
+def generate_spinbox(option: settngs.Setting, layout: QtWidgets.QGridLayout) -> QtWidgets.QSpinBox:
+    row = layout.rowCount()
+    lbl = QtWidgets.QLabel(option.display_name)
```
lordwelch, line 85: Labels should get the tooltip as well

comictaggerlib/ui/talkeruigenerator.py
mizaki: I'd already started on something similar and so left the test API string return stuff in. I can take it out if desired?

comictaggerlib/ui/talkeruigenerator.py
lordwelch: This is still missing on some of them

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,225 @@
...
+    """
+
+    # Store all widgets as to allow easier access to their values vs. using findChildren etc. on the tab widget
+    sources: dict = {"tabs": {}}
+
+    # Tab comes with a QVBoxLayout
+    comic_talker_tab_layout = comic_talker_tab.layout()
+
+    talker_layout = QtWidgets.QGridLayout()
+    lbl_select_talker = QtWidgets.QLabel("Metadata Sources:")
```
lordwelch: This needs to indicate that it is the currently selected talker at a minimum change it back to metadata source

```suggestion
    lbl_select_talker = QtWidgets.QLabel("Metadata Source:")
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -0,0 +1,225 @@
...
+    # Add combobox to sources for getting and setting talker
+    sources["cbx_select_talker"] = cbx_select_talker
+
+    # Add source sub tabs to Comic Sources tab
+    for talker_id, talker_obj in talkers.items():
+        # Add source to general tab dropdown list
+        cbx_select_talker.addItem(talker_obj.name, talker_id)
+
+        tab_name = talker_id
+        sources["tabs"][tab_name] = {"tab": QtWidgets.QWidget(), "widgets": {}}
```
lordwelch: This should be a named tuple instead
```python
class TalkerTab(NamedTuple):
    tab: QtWidgets.QWidget
    widgets: dict[str, QtWidgets.QWidget]
```

comictalker/talkers/comicvine.py
```diff
@@ -203,7 +216,11 @@ def parse_settings(self, settings: dict[str, Any]) -> dict[str, Any]:
         self.remove_html_tables = settings["cv_remove_html_tables"]
         return settings
 
-    def check_api_key(self, key: str, url: str) -> bool:
+    def check_api_key(
+        self,
+        url: str,
+        key: str,
+    ) -> str:
```
lordwelch: Is there a reason for this to be expanded?
```suggestion
    def check_api_key(self, url: str, key: str) -> str:
```

comictaggerlib/ui/talkeruigenerator.py
lordwelch: Make it return a `tuple[str, bool]`

lordwelch (approved):
When these last few comments are addressed I'll merge this.

mizaki:
The final "Make it return a `tuple[str, bool]`" didn't make sense to me for line 201?

lordwelch:
> The final "Make it return a `tuple[str, bool]`" didn't make sense to me for line 201?

Right now it is only used to give a message in the GUI, but if it is ever to be used elsewhere I need to be able to determine if it is successful or not. With only a string I cannot determine if the test was successful or not.

lordwelch:
Now this extra commit needs to be removed aba59bdbfe6aea0a2ef1f9757dd0406bffc84a6d and I can merge

mizaki:
> > The final "Make it return a `tuple[str, bool]`" didn't make sense to me for line 201?
> 
> Right now it is only used to give a message in the GUI, but if it is ever to be used elsewhere I need to be able to determine if it is successful or not. With only a string I cannot determine if the test was successful or not.

The one for the API test I understood. It was the same comment for
```
for option in config[1][f"talker_{talker_id}"][1].values():
            current_widget = None
```
that had me confused. I wondered if it was a c/p mistake from the API one?

mizaki:
> Now this extra commit needs to be removed [aba59bd](https://github.com/comictagger/comictagger/commit/aba59bdbfe6aea0a2ef1f9757dd0406bffc84a6d) and I can merge

Not sure how that got in there. Must have been from the merge somehow. Reverted it.

lordwelch:
> I wondered if it was a c/p mistake from the API one?

No idea how it got there???

Issue #234, opened 2022-03-27
Populate series name only


rhaussmann:
When a series is found only by modifying the series name but the newest issue is not populated yet, allow the option to copy over the series title only to the file.

lordwelch:
This is a good idea, unfortunately ComicTagger is designed to work primarily at the issue level, both for metadata and for determining which series it will get matched with, there is not have a simple, straightforward way to add it at this time.

PR #431, opened 2023-02-15
Support talkers without issue level details

mizaki:
I believe this is where we landed on the non-issue containing talkers. A talker without issues is still expected to return a list from `fetch_issues_by_series_issue_num_and_year`.

### DIFF
--- a/comictaggerlib/issueidentifier.py
+++ b/comictaggerlib/issueidentifier.py
@@ -18,7 +18,7 @@ from __future__ import annotations
 import io
 import logging
 import sys
-from typing import Any, Callable
+from typing import Any, Callable, cast
 
 import settngs
 from typing_extensions import NotRequired, TypedDict
@@ -296,7 +296,7 @@ class IssueIdentifier:
                 primary_img_url, blocking=True
             )
         except ImageFetcherException as e:
-            self.log_msg("Network issue while fetching cover image from Comic Vine. Aborting...")
+            self.log_msg(f"Network issue while fetching cover image from {self.talker.name}. Aborting...")
             raise IssueIdentifierNetworkError from e
 
         if self.cancel:
@@ -318,7 +318,7 @@ class IssueIdentifier:
                         alt_url, blocking=True
                     )
                 except ImageFetcherException as e:
-                    self.log_msg("Network issue while fetching alt. cover image from Comic Vine. Aborting...")
+                    self.log_msg(f"Network issue while fetching alt. cover image from {self.talker.name}. Aborting...")
                     raise IssueIdentifierNetworkError from e
 
                 if self.cancel:
@@ -393,13 +393,20 @@ class IssueIdentifier:
         keys["issue_number"] = IssueString(keys["issue_number"]).as_string()
 
         # we need, at minimum, a series and issue number
-        if not (keys["series"] and keys["issue_number"]):
-            self.log_msg("Not enough info for a search!")
-            return []
+        # Unless this is a series only talker
+        if self.talker.has_issues:
+            if not (keys["series"] and keys["issue_number"]):
+                self.log_msg("Not enough info for a search!")
+                return []
+        else:
+            if not keys["series"]:
+                self.log_msg("Not enough info for a search!")
+                return []
 
         self.log_msg("Going to search for:")
         self.log_msg("\tSeries: " + keys["series"])
-        self.log_msg("\tIssue:  " + keys["issue_number"])
+        if keys["issue_number"]:
+            self.log_msg("\tIssue:  " + keys["issue_number"])
         if keys["issue_count"] is not None:
             self.log_msg("\tCount:  " + str(keys["issue_count"]))
         if keys["year"] is not None:
@@ -429,7 +436,7 @@ class IssueIdentifier:
 
             # remove any series that starts after the issue year
             if keys["year"] is not None and item.start_year is not None:
-                if keys["year"] < item.start_year:
+                if int(keys["year"]) < int(item.start_year):
                     date_approved = False
 
             for name in [item.name, *item.aliases]:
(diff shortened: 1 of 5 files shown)


lordwelch:
I don't see the point in this change, what file are you trying to tag that needs this and what source needs this?
Here is a minimum example with MangaUpdates https://github.com/lordwelch/comictagger/tree/MangaUpdates what does making this change add?

mizaki:
The TL;DR is to block the issue window and prevent the `ii` search of the issue window if the talker does not have issues. And to allow `ii` to work for auto tagging if there is no issue number, which a series only talker won't have.

We had some discussion about it all in https://github.com/comictagger/comictagger/pull/403. The easiest way to get `ii` to work with minimal changes is the "fake" series information as issue information. I did it the "hard" way to start with by making large changes to `ii` that would remove the requirement of the `fetch_issues_by_series_issue_num_and_year`. I also suggested that `fetch_issues_by_series_issue_num_and_year` could be used to "fake" the issue information too and that is what you seemed to prefer going by you adding that function to my draft MU talker. It does seem a much simpler way.

I have my MU talker mostly ready but it requires the other pending PRs. I can put it up as a draft though. It isn't really different in the `fetch_issues_by_series_issue_num_and_year` than yours.

Am I missing something? From my testing `ii` won't work for auto tag without an issue number and one shouldn't be a requirement for a series only talker?

lordwelch:
Instead of making an arbitrary separation between `Talkers` that support issues and talkers that do not support issues. The `IssueIdentifier` should be updated to not require an issue number.

There are no comics or manga that do not have an issue (or with manga chapter/volume) number, if they don't have multiple issues then it is implicitly '1'.
The same goes for any source of metadata, it may not know how many issues there are but there is always at least 1 issue (for american comics there is not a guarantee that it starts with '1', but that is very rare and at least on comicvine is noted as such).

So instead of trying to make a completely separate type of `Talker` that only returns the volume information, keep all of the talkers the same and then only use the Volume information.

mizaki:
> Instead of making an arbitrary separation between `Talkers` that support issues and talkers that do not support issues. The `IssueIdentifier` should be updated to not require an issue number.
> 
> There are no comics or manga that do not have an issue (or with manga chapter/volume) number, if they don't have multiple issues then it is implicitly '1'. The same goes for any source of metadata, it may not know how many issues there are but there is always at least 1 issue (for american comics there is not a guarantee that it starts with '1', but that is very rare and at least on comicvine is noted as such).
> 
> So instead of trying to make a completely separate type of `Talker` that only returns the volume information, keep all of the talkers the same and then only use the Volume information.

I'm wondering then how to know if a talker has issues? If there is an issue number we can presume nothing about the talker. It could be that `fetch_issues_by_series_issue_num_and_year` is called and returns "NotImplemented" and so it is a talker without issues. That is no better really than having the class var `has_issues`? Similar with a function to return `True` or `False`.

Similar problem for the issues window. Other than trying to list the issues and failing, isn't it better to know before hand and not look and not enable the buttons?

Checking the series covers also doesn't make sense for a talker with issues, whereas, those without it might (possibly try if easier checks fail).

I may be missing something obvious here?

An `ii` rework/refinement might be a good idea in general to cover the "new world" with more talkers.

lordwelch:
My entire point is that it's not up to a Talker to do anything different. It is up to the IssueIdentifier to request the data from the Talker and to then figure out what data applies to the current comic it is searching for.
Right now the minimum required information to complete an auto-complete search is:
- Series name
- Issue Number

Everything else is just a nice to have.
We have two problems that need to be solved:
- #234
- Manga websites that don't have at least an issue count

If we simply make the assumption that all series that have been published have to have 1 issue (mainly because it's really hard to publish a series with no content) then all functions on the talker will always return successfully, it may be asked for issue 12 of a series that only has 1 issue but now we are back to only having one problem:
- #234

mizaki:
`Issue window`
The issue window problem first. Can we agree that disabling the button makes sense before someone clicks it? If so, the series window needs to know to disable the button. I see these possible ways:
- 1. Some var on the talker.
- 2. A function that returns True or False.
- 3. Making a call to an issue function and getting NotImplemented (this is the same as 2 really).

Issue window alone, do you prefer any of those or have another suggestion?

`Issue Identifier`
Here's some info on a few APIs. Does the talker have issue/volume numbers returned by the search API? I'm going to use CV, MU, jikan (https://jikan.moe) and kitsu (https://kitsu.io) here.

CV: Has issue count and issue data.

MU: No issue count via search but can get a number with a full info call, it may be issues, it may be chapters, it maybe issues called volumes. It does **not** have issue level data.

kitsu: It does have volumes and issue numbers BUT only after the series has finished (it would appear going by naruto having them and One Piece not). It **does** have issue/chapter data (not usually a lot, if any) BUT does **not** have issue covers.

jikan: Only has issue and volume number after manga is finished. **No** issue data.

So, if `ii` was to assume a missing issue number was 1:
CV: It could check the cover of issue 1. If there is no match and there is no other data, should it fail or go through every cover (for 2000 AD, that's over 2300 issues)? This could be made an option? It could try every issue with a 1 in it to match the cover. How much of an attempt should be made to filter the list (like is done with year currently)? Should the end case be when all other options have been tried, tag with the series data? Could be made an option?

MU: It will have no issue number in the ComicIssue format of the full series data (as draft). We could presume then that if there is no issue number in the data from `fetch_issues_by_series_issue_num_and_year` that it is a series only talker and therefore `ii` should use the series data in the ComicIssue. It could try the covers for a match too but that is a low chance of working. It also possible there is only one result and so that could be a factor, again, using the empty issue number as the deciding factor. Could be made an option?

kitsu: This looks like it'll be an odd one where the talker will need to mix the series data in due to limited issue data. The only thing for this is the issue number. So if 1 was assumed it would be tagged with the wrong title if it wasn't issue/chapter 1.

jikan: Will end up much like MU, passing in ComicIssue with no issue number.

All that said (probably more than needed), `ii` could work with the assumption of 1 for the issue and then depending on what comes back from `fetch_issues_by_series_issue_num_and_year` make choices from there.

I'm not sure if any of that makes sense?

mizaki:
Going forward this can be split into two; the GUI issues window and Issue Identifier without issue number. The issues window is the more important for finalising the multi-talker support imo. Before I can do a PR for that I need to know which, if any of the possibilities to do.

PR #433, opened 2023-02-15
Add additional fields

mizaki:
This adds genres, tags, rating and manga to ComicIssue and adds genres field to ComicSeries to allow for filtering of search results from the cache.


mizaki:
I've added a check for manga and rating. Not sure if it's worth doing the manga? With the rating, should fixes it to 2 decimal points be done?

mizaki:
The MacOS build seems to have stalled. Does it need a reset?

lordwelch:
They finally got rid of macOS 10.15

mizaki:
https://github.com/comictagger/comictagger/pull/433/commits/d65a120eb5b00c594a43e4f7738e30af9ad89ad0 Adds issue_count from count_of_issues. I believe we talked about this before and couldn't rely on CV to have a valid number. I believe it should be up to the talker to validate this number. If the series has ended or been cancelled then the number can be considered valid for example.

mizaki:
With additional talkers comes additional information. Should everything in metadata (that makes sense) be added the ComicSeries and ComicIssue types? It looks like these are the ones left:
language
country
maturity_rating

CoMet-only items (ignore these?)
    price
    is_version_of # Not sure what this is?
    rights # Copyright holder? Same as publisher is so.
    identifier # ISBN etc.?
    last_mark # No idea?

lordwelch:
> language
> country
> maturity_rating

you can just do these three

mizaki:
> > language
> > country
> > maturity_rating
> 
> you can just do these three

Looking at GCD they put the language and country IDs in the series table. I was thinking that putting it in the issue level makes more sense but now I'm not so sure. Country will come from the publisher but a publisher could publish in multiple languages? I'm probably over thinking it as it gets thrown to metadata and sits with the issue anyway.

Any strong opinion on series or issue for language or country attachment?

lordwelch:
> Any strong opinion on series or issue for language or country attachment?

Put it in with the issue, it could be stored on either one but issues are the only option that is guaranteed to be correct every time. for example ComicVine separates the Japanese manga from it's english (or other translation) into a separate series, MangaUpdates has a separate place where they mention the english publisher.

mizaki:
When it comes to checking/converting the language and country I think it would be best if `talker_utils` handled it. Taking either the 2-letter code (and checking it against pycountry) or the language/country name and converting it to the 2-letter code. Either `map_comic_issue_to_metadata` handles it or add some functions to talker_utils (this happens either way, it's whether the talkers are expected to use them or not.)

Do you agree `talker_utils` is the best place and/or when `map_comic_issue_to_metadata` processes the comic data?

Some info on sources:

- MU the publisher type field is language when "original" is returned for the publisher (this can be Japanese, Korean etc.) (https://www.mangaupdates.com/publisher/91qzdxd/viz). 
- From the GCD DB, both 2 letter code and full name is available for language and country of the series. (https://docs.comics.org/wiki/Current_Schema)
- CV has the country on the web page of the publisher but does **not** provide it via the API (back to scrapping the page like used to do with alt images if we want the data). (https://comicvine.gamespot.com/api/documentation#toc-0-26 vs https://comicvine.gamespot.com/viz/4010-428/)

In general it seems an additional API call will be required to infer either country or language so I plan to make it an option with the requisite help message.

mizaki:
I've noticed now that `language` is 2-letter country code and `country` is the name. This is to do with them being ComicInfo and CBL respectively. Annoying but there go. 

I'm wondering for down the line (metadata plugins) if both should be in the generic metadata:
- languageISO
- language_name
- countryISO
- country_name

That way the exporter can choose?

mizaki:
I also missed format which I've now added.

It might make sense to add to `utils` a function to return `format/manga/maturity rating` list so they can be checked against in `talker_utils` when processing the `ComicIssue/ComicSeries` to `genericmetadata`. That is probably better for another PR?

Also adding `countryISO` and renaming language to `languageISO` in `genericmetadata` (add on on to store date PR with whatever new fields are needed also?).

comicapi/utils.py
```diff
@@ -239,6 +239,7 @@ def get_language_from_iso(iso: str | None) -> str | None:
 def get_language_iso(string: str | None) -> str | None:
     if string is None:
         return None
+    # Why return this if ISO failed?
     lang = string.casefold()
```
lordwelch, line 243: This is intended to be a best effort not authoritative

comictalker/talker_utils.py
```diff
@@ -81,6 +91,37 @@ def map_comic_issue_to_metadata(
...
+    # ISO country name
+    if issue_results.country:
+        # Only CBI supports country currently in CT and is a string name
+        # TODO Use https://codeberg.org/plotski/countryguess as pycountry seems stale?
+        # Return current entry if conversion fails
+        metadata.country = issue_results.country
+        if not utils.get_country_iso(issue_results.country):
+            code = utils.get_country_iso_name(issue_results.country)
+            if code:
+                metadata.country = code
```
lordwelch: Put all of this extra validation in a separate PR. Currently the only validation that happens in this function is based on types eg count is an int

comictalker/comiccacher.py
```diff
@@ -117,6 +121,13 @@ def create_cache_db(self) -> None:
                 + "credits TEXT,"  # JSON: "{"name": "Bob Shakespeare", "role": "Writer"}"
                 + "teams TEXT,"  # Newline separated
                 + "story_arcs TEXT,"  # Newline separated
+                + "genres TEXT,"  # Newline separated
+                + "tags TEXT,"  # Newline separated
+                + "critical_rating FLOAT,"
+                + "manga TEXT,"  # Yes/YesAndRightToLeft/No
+                + "maturity_rating TEXT,"
+                + "language TEXT,"  # 2-letter iso code
+                + "country TEXT,"  # 2-letter iso code
```
lordwelch: I don't intend to enforce the format of these. The only enforcement will be data types.

comictalker/talkers/comicvine.py
```diff
@@ -560,6 +564,9 @@ def _format_issue_results(self, issue_results: list[CVIssue], complete: bool = F
 
             series = self._fetch_series_data(record["volume"]["id"])
 
+            # Make below an option?
+            # TODO Infer country by publisher (and language)? Check series desc for "Translated into English"?
```
lordwelch: Keep arbitrary TODOs out of pull requests

lordwelch (changes requested):
format and maturity rating has always been free-form, there will be no validating of these in ComicTagger. A metadata may sometime in the future provide some sort of validation.

mizaki:
Ubuntu build test machine failed getting a package. Might be temporary?

comictalker/resulttypes.py
```diff
@@ -33,7 +37,14 @@ class ComicIssue:
...
     image_url: str
     issue_number: str
+    critical_rating: float
+    maturity_rating: str
+    manga: str
+    genres: list[str]
+    tags: list[str]
     name: str
+    language: str  # 2-letter iso code
+    country: str  # 2-letter iso code
```
lordwelch: Same as before only the type will be enforced here. This should only exist between the cache and a single talker

comictalker/talker_utils.py
```diff
@@ -45,6 +45,16 @@ def map_comic_issue_to_metadata(
     metadata.series = utils.xlate(issue_results.series.name)
     metadata.issue = IssueString(issue_results.issue_number).as_string()
 
+    # Rely on comic talker to validate this number
+    metadata.issue_count = utils.xlate_int(issue_results.series.volume)
+
+    # TODO Generate list in utils (same as language and countries) and check against that in separate PR
```
lordwelch: As in my other comment format is a free-form field with some suggested defaults it will probably never be enforced

lordwelch:
> Ubuntu build test machine failed getting a package. Might be temporary?

That's because I need to tell it to fetch the latest package list before telling it to install a package. That's one reason why I hate the apt package manager.

comictalker/resulttypes.py
mizaki: I suppose really this should be validated by the outputting module (CBI, ComicInfo, etc.).

comicapi/utils.py
```diff
@@ -248,6 +249,39 @@ def get_language_iso(string: str | None) -> str | None:
...
+        return None
+
+    try:
+        country = pycountry.countries.search_fuzzy(string)
+        if country:
+            # Assume first entry is correct
+            return country[0].name
+    except LookupError:
+        pass
+    return None
```
mizaki: Do you want this removed as well?

mizaki:
Other than my outstanding question about removing `get_country_iso` and `get_country_iso_name` I believe I've covered everything?

comicapi/utils.py
lordwelch: Yes lets remove these as well

lordwelch (approved):

PR #437, opened 2023-02-18
Mangaupdates talker

mizaki:
Not tested yet. Reliant on some of the other PRs.


mizaki:
I've added support for `fetch_issues_by_series` but all is does is duplicate the series selected and so there will only ever be one "issue" listed. See PR #471 for a solution as I think showing the issue list with the same information and only one issue will be a bad user experience (and doesn't make a lot of sense).

PR #440, opened 2023-02-25
Feature/add appimage support

bmfrosty:
Changes:
* Added the libfuse2 linux dependency that is required by AppImage
* Added additional files needed to build the AppImage
* Updated the makefile to install appimagetool
* Added command to build the appimage
* Updated build.yaml to also upload .AppImage files

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -89,7 +89,7 @@ jobs:
         if: runner.os == 'macOS'
       - name: Install linux dependencies
         run: |
-          sudo apt-get install pkg-config libicu-dev libqt5gui5
+          sudo apt-get install pkg-config libicu-dev libqt5gui5 libfuse2
           export PKG_CONFIG_PATH="/usr/local/opt/icu4c/lib/pkgconfig";
           export PATH="/usr/local/opt/icu4c/bin:/usr/local/opt/icu4c/sbin:$PATH"
           python -m pip install --no-binary=pyicu pyicu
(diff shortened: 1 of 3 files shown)


lordwelch:
I have a few things I want changed
- Remove the -m from zip (that's the flag that deletes files) so that the PyInstaller build can be re-used for the appimage
- Add dist as a dependency for appimage `appimage:` to `appimage: dist` and remove the PyInstaller command in the appimage target
- Copy 'cp -a' the dist/comictagger folder to build/appimage (I don't want to modify build files in-place, and the folder name does not matter to the result in this case)
- You want to use the [latest release] for the appimage tool not the continuous build which goes off of the latest git commit.
- Instead of making an AppRun script create a symlink to comictagger
- Instead of creating an AppRun.desktop use desktop-integration/linux/ComicTagger.desktop as a template e.g. `sed -e 's|/usr/local/share/comictagger/app.png|app|g' -e 's|%%CTSCRIPT%% %F|./comictagger|g'`

If you could also [squash your commits] and force-push to your branch to make this a single change that would be appreciated.

[latest release]: https://github.com/AppImage/AppImageKit/releases/latest/download/appimagetool-x86_64.AppImage
[squash your commits]: https://stackoverflow.com/a/5201642

bmfrosty:
That's all seems reasonable.    

Only one I'm wary of is the .destop file as appimagetool is a bit picky.  I'll find out.

I'm going to close the PR, do all my work in my own fork.

bmfrosty:
reopening and then catching my changes before I squash.  I'm *so good* at screwing up manual squashes.

bmfrosty:
Squashed.

PR #442, opened 2023-03-03
FIX: Do not fail when talker key or URL is missing when set to False

mizaki:
If `parser.add_setting(f"--{self.id}-key", file=False, cmdline=False)` to used by a talker because it doesn't require a key a `KeyError` will occur and the talker is then not available.

### DIFF
--- a/comictalker/comictalker.py
+++ b/comictalker/comictalker.py
@@ -132,7 +132,7 @@ class ComicTalker:
         """
         if settings.get(f"{self.id}_key"):
             self.api_key = settings[f"{self.id}_key"]
-        if settings[f"{self.id}_url"]:
+        if settings.get(f"{self.id}_url"):
             self.api_url = fix_url(settings[f"{self.id}_url"])
 
         settings[f"{self.id}_url"] = self.api_url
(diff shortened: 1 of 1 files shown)


comictalker/comictalker.py
```diff
@@ -130,7 +130,7 @@ def parse_settings(self, settings: dict[str, Any]) -> dict[str, Any]:
         settings is a dictionary of settings defined in register_settings.
         It is only guaranteed that the settings defined in register_settings will be present.
         """
-        if settings[f"{self.id}_key"]:
+        if settings.get(f"{self.id}_key"):
             self.api_key = settings[f"{self.id}_key"]
         if settings[f"{self.id}_url"]:
```
lordwelch: Lets add the same fix here

mizaki: Added URL. My thinking was the URL is always needed but it is possible the talker wouldn't want to expose the option?

lordwelch: Someone could come up with something that doesn't use a url. Does it make sense? No. but the world doesn't make sense and this is a 5 character fix that doesn't change the standard usecase.

mizaki: I suppose you could do something with files/sqlite etc... Anyway added it in https://github.com/comictagger/comictagger/pull/442/commits/4218e3558bb7577d72ebd37e0ac36ec6d4e55e84

mizaki: Being able to not have a url does indeed make sense as I'm throwing together a talker to use the GCD dump (converted to sqlite) right now.

PR #465, opened 2023-05-13
Hide the API key field text as a password and add a show/hide button

mizaki:
I did wonder about renaming "{talker}-key" to "{talker}-token" to try and cover key/token/password. display_name and help can be altered so probably not needed. The only place it may be confusing a little is on the command line:

```
talker_metron:
  --metron-key METRON_KEY
                        Use the given Metron API password. (default: )
```

### DIFF
--- a/comictaggerlib/ui/talkeruigenerator.py
+++ b/comictaggerlib/ui/talkeruigenerator.py
@@ -40,6 +40,14 @@ def generate_api_widgets(
         else:
             QtWidgets.QMessageBox.warning(None, "API Test Failed", check_text)
 
+    def show_key(le_key: QtWidgets.QLineEdit) -> None:
+        current_state = le_key.echoMode()
+
+        if current_state == 0:
+            le_key.setEchoMode(QtWidgets.QLineEdit.EchoMode.PasswordEchoOnEdit)
+        else:
+            le_key.setEchoMode(QtWidgets.QLineEdit.EchoMode.Normal)
+
     # get the actual config objects in case they have overwritten the default
     talker_key = config[1][f"talker_{talker_id}"][1][f"{talker_id}_key"]
     talker_url = config[1][f"talker_{talker_id}"][1][f"{talker_id}_url"]
@@ -50,8 +58,9 @@ def generate_api_widgets(
     # only file settings are saved
     if talker_key.file:
         # record the current row so we know where to add the button
-        btn_test_row = layout.rowCount()
+        btn_show_key = layout.rowCount()
         le_key = generate_textbox(talker_key, layout)
+        le_key.setEchoMode(QtWidgets.QLineEdit.EchoMode.PasswordEchoOnEdit)
         # To enable setting and getting
         sources["tabs"][talker_id].widgets[f"talker_{talker_id}_{talker_id}_key"] = le_key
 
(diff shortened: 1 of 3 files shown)


comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -64,6 +73,12 @@ def call_check_api(*args: Any, le_url: QtWidgets.QLineEdit, le_key: QtWidgets.QL
...
         sources["tabs"][talker_id].widgets[f"talker_{talker_id}_{talker_id}_url"] = le_url
 
+    # The key button row was recorded so we add the show/hide button
+    if btn_show_key is not None:
+        btn_show = QtWidgets.QPushButton("Show/Hide")
+        layout.addWidget(btn_show, btn_show_key, 2)
+        btn_show.clicked.connect(partial(show_key, le_key=le_key))
+
     # The button row was recorded so we add it
     if btn_test_row is not None:
```
lordwelch: You renamed btn_test_row and added a new check but left the old one in?

mizaki: No. It's a new button to show and hide the `key` field which is now a password by default.

image

Should I alter the comment to make it clearer?

lordwelch: This now no longer has a `Test API` button if the url setting is disabled `talker_url.file == False` and the key setting is enabled `talker_key.file == True` where as before it still had the button. This will also create an exception if `talker_key.file == False`

mizaki: Okay, yes. I had forgotten we'd talked about this before and how the test button works. I think the best solution will then be this way: https://kushaldas.in/posts/creating-password-input-widget-in-pyqt.html

#442 is to address the problem of a missing key or url.

lordwelch: Yes that will be better

mizaki: I used an 80% grey icon to try and cover light and dark. I used the icons from https://github.com/pythonguis/python-qtwidgets which should be fine?

lordwelch: Yes, they use the MIT License which is fine.

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -58,9 +64,16 @@ def show_key(le_key: QtWidgets.QLineEdit) -> None:
...
+        btn_test_row = layout.rowCount()
         le_key = generate_textbox(talker_key, layout)
+
+        # Generate show/hide icon for key/password within the text box
         le_key.setEchoMode(QtWidgets.QLineEdit.EchoMode.PasswordEchoOnEdit)
+
+        le_key_toggle = le_key.addAction(visibleIcon, QtWidgets.QLineEdit.TrailingPosition)
+        le_key_toggle.setToolTip("Show/Hide")
+        le_key_toggle.triggered.connect(partial(show_key, le_key=le_key, le_action=le_key_toggle))
+
```
lordwelch: > I think the best solution will then be this way: https://kushaldas.in/posts/creating-password-input-widget-in-pyqt.html

There should be a new `generate_password_textbox` that uses a class similar to the one created in the above link instead of patching this in after the fact.

mizaki: I copied it wholesale in https://github.com/comictagger/comictagger/pull/465/commits/057725c5da0e7c9032cf5985d8bcaab129431182

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -118,6 +120,51 @@ def generate_textbox(option: settngs.Setting, layout: QtWidgets.QGridLayout) ->
     return widget
 
 
+def generate_password_textbox(option: settngs.Setting, layout: QtWidgets.QGridLayout) -> QtWidgets.QLineEdit:
+    class PasswordEdit(QtWidgets.QLineEdit):
```
lordwelch: This should be a top level class declaration it should not be in the function

PR #466, opened 2023-05-14
Attribution for metadata provider on command line

mizaki:
Adds the talker used for searching and adds attribution line after tagging.

### DIFF
--- a/comictaggerlib/cli.py
+++ b/comictaggerlib/cli.py
@@ -82,8 +82,6 @@ class CLI:
                 logger.info("dry-run option was set, so nothing was written, but here is the final set of tags:")
                 print("dry-run option was set, so nothing was written, but here is the final set of tags:")
                 print(f"{md}")
-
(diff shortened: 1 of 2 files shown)


comictaggerlib/cli.py
```diff
@@ -82,6 +82,8 @@ def actual_metadata_save(self, ca: ComicArchive, md: GenericMetadata) -> bool:
                 logger.info("dry-run option was set, so nothing was written, but here is the final set of tags:")
                 print("dry-run option was set, so nothing was written, but here is the final set of tags:")
                 print(f"{md}")
+
+        print(f"Metadata provided by {self.current_talker().name} {self.current_talker().website}")
```
lordwelch: This is not a good place for this. This will show for every comic that is saved. This can be listed once when the CLI is started or once when it is ended.

mizaki:
I've moved it to after the `post_process_matches` display. That should be better?

PR #467, opened 2023-05-28
Call new talker function on item select

mizaki:
This would allow better support for talkers lacking all the information required for the GUI in their "list" API calls. Such as not returning the image URL, description, publisher, etc.

It may be advantageous to add a "complete" to the series table too.


lordwelch:
In general the Talker should just make the extra call to the api.

mizaki:
> In general the Talker should just make the extra call to the api.

If you mean at search or series list time, that can result in a lot (hundreds, possibly thousands) of unneeded extra calls and extended waits times into the minutes. It's the same as loading all the cover images before they are required. To be respectful to the API, fetching additional data as needed (for a small list with auto-tagging or when selected in the GUI) seems like a sane approach.

mizaki:
I've added a `complete` field to ComicSeries. This will prevent the possibly of every time a series is selected in the GUI that a new fetch from the API is triggered.

I'm not sure there is another way around doing something like this for the GUI when supporting multiple APIs. The only other option I see is limiting the results for series and issues and then having `next`/`previous` buttons. 

An extreme example of having to make thousands of additional API calls would be 2000 AD which has 2335 issues currently.

lordwelch:
What is the difference between using these new functions to get the individual series and issue info and calling the existing functions that do the same thing?

mizaki:
I'll give a couple of examples:

Metron:
`search_for_series` is run (GUI and CLI) and will not return an image for the series. To get an image (usually the first issue) would require an additional call to list all the issues. That would be 1 extra call for each item in the search result (which is capped at 500) that a user may never select and see.

`fetch_issues_by_series` (GUI only function) fetches the issue list but the API call does not return the issue title or the description. A call for the full issue information will be required (and the talker should cache it too so if that is selected for tagging no new API call is required). Again, this could be up to thousands of calls for the descriptions that may never be needed.

MangaUpdates:
The publisher is missing from search and requires another call for each series item to populate.

GCD (Not an API but still represents the issue)
The issue list is missing the image URL. For each issue in the issue list a single scrape is required. A single scrape takes about 1-2 seconds and so would add up quickly.

When it comes to `issueidentifer` which uses `search_for_series` and `fetch_issues_by_series_issue_num_and_year` the extra calls aren't required. For `search_for_series` as it's only a title and `fetch_issues_by_series_issue_num_and_year` will have to do these additional calls but shouldn't be as much of a problem as the number required will be limited via the issue number. There is also no way around the extra calls as the information is required for `ii` to work.

Hopefully that makes it clearer that the additional functions will not do the same as any of the current functions. In a general sense all "search" and "issue list" API calls return a limited subset of information and a single API to a different endpoint returns all available information which may be wanted for the GUI (CV excluded because CT was built around its information). Taking into account possible API call limits (just 1 every second for example) and delays of many minutes for information that the user will never require would not be out of the norm.

An argument about how much it matters if there is missing information can be made but then you could say the same about loading the cover images etc. An issue list with only the issue number and nothing else is not a great user experience and people using the GUI you would imagine want to see the information. Although only filling it in when someone clicks on an item isn't the best, it seems like the best middle ground.

lordwelch:
So that's not the direction I was trying to go in, what I meant is what information does `fetch_additional_issue_info_on_click` provide that `fetch_comic_data` doesn't and why wouldn't we just add that information to `fetch_comic_data`

mizaki:
> So that's not the direction I was trying to go in, what I meant is what information does `fetch_additional_issue_info_on_click` provide that `fetch_comic_data` doesn't and why wouldn't we just add that information to `fetch_comic_data`

Should be with you now. `fetch_comic_data` returns `GenericMetadata`. A similar function `fetch_full_comic_data` to return ComicSeries | ComicIssue and do so based on `series_id` and `issue_id` could be added:
```
    def fetch_full_comic_data(
        self, issue_id: str | None = None, series_id: str | None = None
    ) -> ComicSeries | ComicIssue | None:
        comic_data = None
        if issue_id:
            comic_data = self._fetch_issue_data_by_issue_id(issue_id)
        elif series_id:
            comic_data = self._fetch_series_data(int(series_id))

        return comic_data
```
The `series_id and issue_number` was left in `fetch_comic_data` as a "maybe" fallback but with how CT works now, there should never be an instance where the `issue_id` is unknown.

Is this more where you intended?

lordwelch:
Ok, so ignoring `ComicSeries` for the moment what does `fetch_full_comic_data` provide that `fetch_comic_data` doesn't and why wouldn't we just add that information to `fetch_comic_data`

mizaki:
> Ok, so ignoring `ComicSeries` for the moment what does `fetch_full_comic_data` provide that `fetch_comic_data` doesn't and why wouldn't we just add that information to `fetch_comic_data`

As mentioned, it returns `GenericMetadata` when ComicSeries/ComicIssue is needed because the current list of ComicSeries/ComicIssue will be updated with the information.

lordwelch:
Ok, so what does `ComicIssue` provide that `GenericMetadata` doesn't and why wouldn't we just add that information to `GenericMetadata`

mizaki:
> Ok, so what does `ComicIssue` provide that `GenericMetadata` doesn't and why wouldn't we just add that information to `GenericMetadata`

I think we both may be getting lost here. 

Everything comic talker related is in `ComicSeries` and `ComicIssues`, going to and from the cache, to `issueidentifer`, the GUI windows etc. until it's ready to be sent out into the `taggerwindow` fields/saved and then it's converted `GenericMetadata`. If `GenericMetadata` was used it would then have to be converted back to `ComicSeries`/`ComicIssue` because the series window and issue window use them respectively. So it would go: `CV` -> `ComicIssue` -> `GenericMetadata` -> `ComicIssue` to update the `list[ComicIssue]` used by the issue window.

I feel like I should be seeing something obvious but I can't? It's the same "data" but formatted in different ways.

PR #469, opened 2023-05-29
Add option to presume the issue number as one

mizaki:
This seems like the simplest solution to the missing issue number problem. If there is no issue number and the option is selected (false by default) then when the archive is loaded, the issue number "1" is added.

### DIFF
--- a/comictaggerlib/ctsettings/file.py
+++ b/comictaggerlib/ctsettings/file.py
@@ -83,6 +83,12 @@ def identifier(parser: settngs.Manager) -> None:
         action=argparse.BooleanOptionalAction,
         help="Clears all existing metadata when applying metadata from comic source",
     )
+    parser.add_setting(
+        "--presume-issue-one",
+        default=False,
+        action=argparse.BooleanOptionalAction,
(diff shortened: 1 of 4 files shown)


lordwelch:
My issue with this is that we already have this option.
```
  -1, --assume-issue-one, --no-assume-issue-one
                        Assume issue number is 1 if not found (relevant for -s).
```
It is restricted to auto-tagging specifically because it is not always correct and if you are in the GUI interactively searching / auto-identifying an issue you can type the number 1 easily enough. Sure it's a little annoying but if you want the fully automated approach that is provided via the auto-tagging feature.

mizaki:
I forgot/missed that was already in there. Blinders on.

I think changing the pop up message with auto-identify would be a good idea then. Right now, if you only read the message `"Can't auto-identify without an issue number (yet!)"` it seems like it's not possible at all. Auto-identify the function vs auto-identify in general (a subset required for auto-tag etc.). Adding a "You can use Auto-tag with 'Assume issue number is 1 if not found' option if desired." make sense?

PR #471, opened 2023-05-30
Disable issues related buttons with series only talkers. GUI only.

mizaki:
This allows the GUI to disable "auto-identify" and "show issues" buttons. I've put an extra property on the talker `has_issue_level_data` which defaults to `True`. With them disabled, double clicking the series is the same as "okay" and will tag with the series data.

No `issueidentifer` changes needed.

Below is a video with it on and off using PR #437.

CT_no_issues
Buttons disabled

CT_issues
Duplicating series data into issue list


mizaki:
Okay, I had another idea that might work: 

If the `count_of_issues` is `0/None` then disable the `Show issues` button in the GUI?

PR #474, opened 2023-06-08
Change auto-identfy message

mizaki:
Point the user to the auto-tag `assume 1` if there is no issue number. It could be thought just by reading the previous message that there was no way to auto-tag without an issue number.

### DIFF
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -1017,7 +1017,10 @@ class TaggerWindow(QtWidgets.QMainWindow):
         # Only need this check is the source has issue level data.
         if autoselect and issue_number == "":
             QtWidgets.QMessageBox.information(
-                self, "Automatic Identify Search", "Can't auto-identify without an issue number (yet!)"
+                self,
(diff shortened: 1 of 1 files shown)


comictaggerlib/taggerwindow.py
```diff
@@ -1017,7 +1017,10 @@ def query_online(self, autoselect: bool = False, literal: bool = False) -> None:
         # Only need this check is the source has issue level data.
         if autoselect and issue_number == "":
             QtWidgets.QMessageBox.information(
-                self, "Automatic Identify Search", "Can't auto-identify without an issue number (yet!)"
+                self,
+                "Automatic Identify Search",
+                "Can't auto-identify without an issue number. The auto-tag function has the 'If no issue number, "
+                'assume "1"\' option if desired.',
```
lordwelch: quotes need to be consistent. running black or `tox r -e format` should fix it

mizaki: I think it was black that broke it after I tried to make it multi-line... Looks okay now.

PR #482, opened 2023-06-20
Add pyrate-limiter and apply CV suggested rate limit

mizaki:
This add pyrate-limiter to the required deps.

I took the (very) old comment from the CV API [forum](https://comicvine.gamespot.com/forums/api-developers-2334/api-rate-limiting-1746419/) and set the rate limit to 1 every 2 seconds. I've left a simple check and 10 second wait as a fallback. Instead of waiting many multiple minutes which might make the (GUI) user think the program has hung, it provides a network error with a link where to check your API usage.

### DIFF
--- a/comictalker/talkers/comicvine.py
+++ b/comictalker/talkers/comicvine.py
@@ -26,6 +26,7 @@ from urllib.parse import urljoin
 
 import requests
 import settngs
+from pyrate_limiter import Limiter, RequestRate
 from typing_extensions import Required, TypedDict
 
 import comictalker.talker_utils as talker_utils
@@ -151,6 +152,10 @@ class CVResult(TypedDict, Generic[T]):
 
 
 CV_STATUS_RATELIMIT = 107
+
+# https://comicvine.gamespot.com/forums/api-developers-2334/api-rate-limiting-1746419/
+# "Space out your requests so AT LEAST one second passes between each and you can make requests all day."
+limiter = Limiter(RequestRate(1, 2))
 
 
 class ComicVineTalker(ComicTalker):
@@ -428,27 +433,29 @@ class ComicVineTalker(ComicTalker):
 
         return formatted_filtered_issues_result
 
+    @limiter.ratelimit("cv", delay=True)
     def _get_cv_content(self, url: str, params: dict[str, Any]) -> CVResult:
         """
-        Get the content from the CV server.  If we're in "wait mode" and status code is a rate limit error
-        sleep for a bit and retry.
+        Get the content from the CV server.  We should never hit a rate limit but will cover it anyway.
         """
-        total_time_waited = 0
-        limit_wait_time = 1
         counter = 0
-        wait_times = [1, 2, 3, 4]
         while True:
             cv_response: CVResult = self._get_url_content(url, params)
-            if self.wait_on_ratelimit and cv_response["status_code"] == CV_STATUS_RATELIMIT:
-                logger.info(f"Rate limit encountered.  Waiting for {limit_wait_time} minutes\n")
-                time.sleep(limit_wait_time * 60)
-                total_time_waited += limit_wait_time
-                limit_wait_time = wait_times[counter]
+
+            if cv_response["status_code"] == CV_STATUS_RATELIMIT:
+                logger.info("Rate limit encountered. Waiting for 10 seconds\n")
+                time.sleep(10)
                 if counter < 3:
                     counter += 1
-                # don't wait much more than 20 minutes
-                if total_time_waited < self.wait_on_ratelimit_time:
                     continue
+                # Tried 3 times, inform user to check CV website.
+                logger.error("Rate limit error. Exceeded 3 retires.")
+                raise TalkerNetworkError(
+                    self.name,
+                    3,
+                    "Rate Limit Error: Check your current API usage limit at https://comicvine.gamespot.com/api/",
+                )
+
             if cv_response["status_code"] != 1:
                 logger.debug(
                     f"{self.name} query failed with error #{cv_response['status_code']}:  [{cv_response['error']}]."
--- a/comictalker/talkers/comicvine.py
+++ b/comictalker/talkers/comicvine.py
@@ -151,6 +151,12 @@ class CVResult(TypedDict, Generic[T]):
     version: str
 
 
+# https://comicvine.gamespot.com/forums/api-developers-2334/api-rate-limiting-1746419/
+# "Space out your requests so AT LEAST one second passes between each and you can make requests all day."
+custom_limiter = Limiter(RequestRate(10, 10))
+default_limiter = Limiter(RequestRate(1, 5))
+
+
 class ComicVineTalker(ComicTalker):
     name: str = "Comic Vine"
     id: str = "comicvine"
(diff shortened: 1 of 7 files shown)


mizaki:
Build error is: E: Failed to fetch http://azure.archive.ubuntu.com/ubuntu/pool/main/m/mesa/libegl-mesa0_22.2.5-0ubuntu0.1%7e22.04.1_amd64.deb  404  Not Found [IP: 40.119.46.219 80]

lordwelch:
So with the current limit of (1, 2) this means that there is at least 2 seconds between every request which more than doubles the time it takes to auto-tag ~50 comics that I tested with. At a minimum this should be (1, 1) but I would rather go with (10, 10). I also want a separate more severe limiter for the shared api key as it can run up api requests faster.

mizaki:
> So with the current limit of (1, 2) this means that there is at least 2 seconds between every request which more than doubles the time it takes to auto-tag ~50 comics that I tested with. At a minimum this should be (1, 1) but I would rather go with (10, 10). I also want a separate more severe limiter for the shared api key as it can run up api requests faster.

I was going by
```
"Space out your requests so AT LEAST one second passes between each and you can make requests all day. Go even a millisecond faster and you'll hit a brick wall REALLY HARD."
```
which doesn't appear to be true. Which I interpreted to mean a gap of 1 second but 1,1 is probably correct. Of course, getting any answer from CV seems very unlikely. Other than the status message, I can't seem to find if the API reports the cool down required (like a HTTP header), going to the website appears to be the only resort...

It also appears it uses the correct [HTTP code](https://comicvine.gamespot.com/forums/api-developers-2334/new-api-http-status-when-rate-limited-1751775/) now for rate limit so I'll switch over to that. 107 is no longer listed as a status code.

It is possible to stack the limit for different [durations](https://github.com/vutran1710/PyrateLimiter#basic-usage). I'm not sure if that's of any benefit here in say a limit per minute and hour instead of seconds.

A option could be added (with a message about the default key being hard coded)? Have a spin box so it can be limited.

mizaki:
I've changed to default to 10,10 and added 1,2 for the default key. What number should the default key be, 1,5 or higher? Putting in some kind of warning about the limit on the default key seems sensible. Only place I can see putting it is in the settings help? Also in the welcome text?

I'm in two minds about making an option for the rate. Could limit it to 1-10 every second?

comictalker/talkers/comicvine.py
```diff
@@ -150,27 +151,24 @@ class CVResult(TypedDict, Generic[T]):
     version: str
 
 
-CV_STATUS_RATELIMIT = 107
```
lordwelch, line 153: Is there a reason why this was removed? It should have been set to 429

comictalker/talkers/comicvine.py
```diff
@@ -151,31 +151,24 @@ class CVResult(TypedDict, Generic[T]):
...
-
-# https://comicvine.gamespot.com/forums/api-developers-2334/api-rate-limiting-1746419/
-# "Space out your requests so AT LEAST one second passes between each and you can make requests all day."
-limiter = Limiter(RequestRate(1, 2))
```
lordwelch: Lets make two limiters here one default limiter and a second custom limiter

comictalker/talkers/comicvine.py
```diff
@@ -151,31 +151,24 @@ class CVResult(TypedDict, Generic[T]):
...
 class ComicVineTalker(ComicTalker):
     name: str = "Comic Vine"
     id: str = "comicvine"
     website: str = "https://comicvine.gamespot.com"
     logo_url: str = f"{website}/a/bundles/comicvinesite/images/logo.png"
     attribution: str = f"Metadata provided by <a href='{website}'>{name}</a>"
 
+    # https://comicvine.gamespot.com/forums/api-developers-2334/api-rate-limiting-1746419/
+    # "Space out your requests so AT LEAST one second passes between each and you can make requests all day."
+    limiter = Limiter(RequestRate(10, 10))
```
lordwelch: This should not be a class attribute move it to __init__ and assign the default limiter from the above comment

comictalker/talkers/comicvine.py
```diff
@@ -217,8 +203,12 @@ def parse_settings(self, settings: dict[str, Any]) -> dict[str, Any]:
         settings = super().parse_settings(settings)
 
         self.use_series_start_as_volume = settings["cv_use_series_start_as_volume"]
-        self.wait_on_ratelimit = settings["cv_wait_on_ratelimit"]
         self.remove_html_tables = settings["cv_remove_html_tables"]
+
+        # Set a different limit if using the default API key
+        if self.api_key == self.default_api_key:
+            ComicVineTalker.limiter = Limiter(RequestRate(1, 2))
```
lordwelch: Here we want to cover both cases otherwise it will not change to the correct limiter every time

comictalker/talkers/comicvine.py
```diff
@@ -436,41 +426,24 @@ def fetch_issues_by_series_issue_num_and_year(
     @limiter.ratelimit("cv", delay=True)
```
lordwelch: When using two different rate limiters we can't use the decorator format you will need to use the context manager https://github.com/vutran1710/PyrateLimiter#contextmanager

lordwelch (changes requested):
Lets set the default limit to (1, 5). Then add to the welcome text that the default ComicVine API Key is rate limited and getting a personal API Key will provide up to 5x speedup for online searches. That should provide sufficient motivation to get their own API Key. We also need to add a link to the api page https://comicvine.gamespot.com/api in the settings so that they know where to get one.

For now lets not make an option. If it proves to be useful in the future it will be simple enough to add later.

> which doesn't appear to be true. Which I interpreted to mean a gap of 1 second but 1,1 is probably correct.

How it works for (x,y) is that the first request marks the beginning of the time and then from that time you get x-1 more requests until y time has passed and then the next request that hasn't expired beyond y is used as the beginning so with (1,1) the first request goes through and starts the timer for 1 second 1-1=0 so you don't have any more requests until 1 second has passed. With (10,10) the first request goes through and starts the timer for 10 seconds 10-1=9 so there are 9 more requests that you can do in the next 10 seconds, if there are 10 more requests made at the 5 second mark, 9 go through without issue the 10th request waits for the remaining 5 seconds until the first request expires.

CV probably won't give any more specific info on it because they don't want people gaming the system

comictalker/talkers/comicvine.py
mizaki: It's no longer listed in the [API doc](https://comicvine.gamespot.com/api/documentation#toc-0-0) and as mentioned, they use the default HTTP code now.

lordwelch: I meant the variable not the value

mizaki: The var isn't used because it's now a HTTP status code like 200, 500, 404 etc.. That value was used after a CV had returned something. Now, it has moved from `_get_cv_content` to `_get_url_content`.

mizaki:
I put a link in the welcome message to the wiki page. (I'm creating a new wiki page for Information Sources too [WIP](https://github.com/comictagger/comictagger/wiki/Comic-Information-Sources)). Because the settings are generated programmatically, I'm not sure on a way to get a pure label in there without making some bigger changes. Something could be added to the tooltip help but obviously isn't highly visible.

comictalker/talkers/comicvine.py
lordwelch: Then please use `requests.status_codes.codes.TOO_MANY_REQUESTS`. Here is the rationale when this was originally added https://github.com/comictagger/comictagger/pulls?q=is%3Apr+is%3Aclosed+talker#discussion_r998881387
None of the numbers should be used be used but that is outside the scope of this PR, and 200, 500 and 404 are the most common and 404 is literally a meme at this point because everyone has used a dead link before. I don't think anyone that has ever used an http library has any doubt as to what those 3 mean; substantially less people know that 429 is the proper HTTP code for too many requests.

lordwelch (approved):

eduo:
Is there logging on the API at all? I'm on mac and checked both the built-in log window and the console but I can't find why I've got an error the last two days that the API access it not valid and no results.

Accessing the web itself works OK. I'd like to see what the API is responding, rather than comivine's interpretation.

I've been doing a ton of searches, refreshing my library's metadata, but for two days it hasn't worked any more.

If I try the API directly via URL I get proper results:

https://comicvine.gamespot.com/api/volumes/?api_key=XXXXX_APIKEY&format=json&sort=name:asc&filter=name:Walking%20Dead

image

lordwelch:
@eduo if you have a problem open an issue. Do not comment on old pull requests.

PR #487, opened 2023-06-26
Add Perceptual Hash computation to imagehasher

mizaki:
This returns the same hex result (from my limit testing) as https://github.com/JohannesBuchner/imagehash/blob/master/imagehash/__init__.py#L260 (phash) but without using scipy or numpy.

### DIFF
--- a/comictaggerlib/imagehasher.py
+++ b/comictaggerlib/imagehasher.py
@@ -17,8 +17,9 @@ from __future__ import annotations
 
 import io
 import logging
+import math
 from functools import reduce
-from typing import TypeVar
+from statistics import median
 
 try:
     from PIL import Image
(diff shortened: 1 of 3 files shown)


comictaggerlib/imagehasher.py
```diff
@@ -71,118 +72,100 @@ def set_bit(x: int, idx_val: tuple[int, int]) -> int:
 
         return result
 
-    def average_hash2(self) -> None:
```
lordwelch: Please return the average_hash2 function it should not be removed in this PR.
Was any of this code taken from another project? If so it needs to be stated here.

comictaggerlib/imagehasher.py
```diff
@@ -71,118 +72,100 @@ def set_bit(x: int, idx_val: tuple[int, int]) -> int:
...
+
+        pixels = convert_image_to_ndarray(image)
+        dct = generate_dct2(generate_dct2(pixels, axis=0), axis=1)
+        dctlowfreq = [row[:8] for row in dct[:8]]
+        med = median([item for sublist in dctlowfreq for item in sublist])
+        # Convert to a bit string
+        diff = "".join(str(int(item > med)) for row in dctlowfreq for item in row)
+        # Convert to hex
+        width = int(math.ceil(len(diff) / 4))
+        result = "{:0>{width}x}".format(int(diff, 2), width=width)
```
lordwelch: A string should not be returned this should be an int.
```python
diff = [int(item > med) for row in dctlowfreq for item in row]
result = int(diff, 2)
```

comictaggerlib/imagehasher.py
```diff
@@ -71,118 +72,100 @@ def set_bit(x: int, idx_val: tuple[int, int]) -> int:
...
+        elif isinstance(h1, str) and isinstance(h2, str):
             # convert hex strings to ints
             n1 = int(h1, 16)
             n2 = int(h2, 16)
+        else:
+            # Mixed hashes or some other problem so return a high number. Should return None instead?
+            return 999
 
         # xor the two numbers
-        n = n1 ^ n2
```
lordwelch: This typing was correct before please revert.

comictaggerlib/issueidentifier.py
```diff
@@ -53,7 +53,7 @@ class SearchKeys(TypedDict):
 class Score(TypedDict):
     score: NotRequired[int]
     url: str
-    hash: int
```
lordwelch: This should always be an int

comictaggerlib/issueidentifier.py
```diff
@@ -132,11 +132,9 @@ def set_hasher_algorithm(self, algo: int) -> None:
     def set_output_function(self, func: Callable[[str], None]) -> None:
         self.output_function = func
 
-    def calculate_hash(self, image_data: bytes) -> int:
-        if self.image_hasher == 3:
-            return -1  # ImageHasher(data=image_data).dct_average_hash()
+    def calculate_hash(self, image_data: bytes) -> int | str:
         if self.image_hasher == 2:
-            return -1  # ImageHasher(data=image_data).average_hash2()
```
lordwelch: Please revert and use `self.image_hasher == 3` for the phash function

comictaggerlib/issueidentifier.py
```diff
@@ -280,7 +278,7 @@ def get_issue_cover_match_score(
         issue_id: str,
         primary_img_url: str,
         alt_urls: list[str],
-        local_cover_hash_list: list[int],
```
lordwelch: This should always be an int

comictaggerlib/resulttypes.py
```diff
@@ -10,7 +10,7 @@ class IssueResult(TypedDict):
     distance: int
     issue_number: str
     cv_issue_count: int | None
-    url_image_hash: int
```
lordwelch: This should always be an int

lordwelch (changes requested):

comictaggerlib/imagehasher.py
mizaki: It was made with the help of ChatGPT. I did a search before that and didn't find anything that I could make fit.

comictaggerlib/imagehasher.py
mizaki: The `diff` change created an array which wouldn't work with `int`. I left `diff` as is to create a bit string and int'ed that.

lordwelch (approved):

PR #488, opened 2023-06-26
Fixes sorting by year and issue count. 

### LINKED ISSUE #475: [Bug] Issue with Sort Functionality - Incorrect Sorting Order
**Description:**
The current sorting functionality in ComicTagger is not working as expected. When attempting to sort a list of numbers, the resulting order is incorrect. Here is an example:

**Actual Sort:**
7
67
62
6

**Expected Sorted List:**
67
62
7
6

Instead, the sorting algorithm seems to be disregarding the numerical value and arranging the numbers in an unexpected manner. 
Pasted Graphic

**Steps to Reproduce:**
1. Open ComicTagger.
2. Select an issue.
3. Search Online.
4. Use the sort function to arrange the list in asc or desc order.

**Expected Behavior:**
The numbers should be sorted in ascending order based on their numerical value.

**Actual Behavior:**
The numbers are sorted in an incorrect order that does not match the expected outcome.

**Additional Information:**
- ComicTagger version: 1.5.5
- Operating System: MacOS Ventura 13.3

mizaki:
Fixes #475

### DIFF
--- a/comictaggerlib/seriesselectionwindow.py
+++ b/comictaggerlib/seriesselectionwindow.py
@@ -464,25 +464,28 @@ class SeriesSelectionWindow(QtWidgets.QDialog):
             item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
             self.twList.setItem(row, 0, item)
 
-            item_text = str(record.start_year)
-            item = QtWidgets.QTableWidgetItem(record.start_year)
-            item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
-            item.setData(QtCore.Qt.ItemDataRole.DisplayRole, record.start_year)
-            item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
-            self.twList.setItem(row, 1, item)
+            if record.start_year is not None:
+                item_text = f"{record.start_year:04}"
+                item = QtWidgets.QTableWidgetItem(item_text)
+                item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
+                item.setData(QtCore.Qt.ItemDataRole.DisplayRole, record.start_year)
+                item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
+                self.twList.setItem(row, 1, item)
 
-            item_text = str(record.count_of_issues)
-            item = QtWidgets.QTableWidgetItem(record.count_of_issues)
-            item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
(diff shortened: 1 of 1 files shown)


comictaggerlib/seriesselectionwindow.py
```diff
@@ -465,24 +465,24 @@ def categorize(result: ComicSeries) -> int:
             self.twList.setItem(row, 0, item)
 
             item_text = str(record.start_year)
-            item = QtWidgets.QTableWidgetItem(item_text)
+            item = QtWidgets.QTableWidgetItem(record.start_year)
```
lordwelch: This sets the item [type] and should not be used instead pad to four characters: `f"{record.start_year:04}"`

[type]: https://doc.qt.io/qt-6/qtablewidgetitem.html#QTableWidgetItem

comictaggerlib/seriesselectionwindow.py
```diff
@@ -465,24 +465,24 @@ def categorize(result: ComicSeries) -> int:
...
             item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
+            item.setData(QtCore.Qt.ItemDataRole.DisplayRole, record.start_year)
             item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
             self.twList.setItem(row, 1, item)
 
             item_text = str(record.count_of_issues)
-            item = QtWidgets.QTableWidgetItem(item_text)
+            item = QtWidgets.QTableWidgetItem(record.count_of_issues)
```
lordwelch: Same as above: This sets the item [type] and should not be used instead pad to four characters: `f"{record.start_year:04}"`

comictaggerlib/seriesselectionwindow.py
```diff
@@ -465,24 +465,24 @@ def categorize(result: ComicSeries) -> int:
...
             item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
             item.setData(QtCore.Qt.ItemDataRole.DisplayRole, record.count_of_issues)
             item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
             self.twList.setItem(row, 2, item)
 
-            if record.publisher is not None:
```
lordwelch, line 480: Unless you can prove that record.publisher is never None this needs to stay there are several other checks to see if the publisher is None in other parts of the code and None of them set it to the empty string.

lordwelch (changes requested):

comictaggerlib/seriesselectionwindow.py
mizaki: If it's `None` then the result is an empty field. It doesn't make sense to do the work though so I'll revert it. I'll also add the same check for `year` and `count_of_issues` as they can be empty too.

lordwelch: I don't remember where exactly but CT has had issues with None as it does not sort with other types `<` or `>` operators in python

lordwelch (approved):

PR #490, opened 2023-07-12
Support translations for the GUI

mizaki:
A few things:
1. For the translations there seems to be two websites: [weblate](https://weblate.org) and [transifex](https://www.transifex.com/open-source/) which are both free for FOSS projects.
2. I've cleaned and consolidated the names between the UI files and class names. Also some other cleaning.
3. A reload of the program is the easiest way to switch languages. Otherwise (afaik) you need to manually re-translate every label.
4. I've included a lang_utils for testing but weblate or transifex will pull the strings for translation in and it appears a GitHub action can be done to generate the .gm files.
5. Some of the lists, such as format/rating/etc. need to be switched to using userData in English so the strings can be translated. (I think this makes sense to do?)
6. I've put in a test en_GB file (`read style` is changed to `read type`) and a custom message when switching to it. I've also included Japanese (but with no files for CT) to show the builtin translations for the widgets works.
7. I've probably missed a string or two somewhere.
8. Because this is using the Qt library, no CLI translations. For that gettext appears to be the answer (and switch the GUI over to it too).
9. How to support plugins? Because settngs is used for the strings for CLI and GUI, it presents a problem. For a string to be pulled by pylupdate for translation it needs to be wrapped in `tr()` which requires Qt. Hoping for a bright idea. Best I have is using `QT_TRANSLATE_NOOP` which would look something like:
```
    def __init__(self, version: str, cache_folder: pathlib.Path):
        super().__init__(version, cache_folder)

        self.remove_html_tables: bool = False
        self.remove_html_tables_txt = QT_TRANSLATE_NOOP("ComicVine", "Remove HTML tables")
        self.remove_html_tables_hlp = QT_TRANSLATE_NOOP("ComicVine", "Removes html tables instead of converting them to text")

    def register_settings(self, parser: settngs.Manager) -> None:
        parser.add_setting(
            "--cv-remove-html-tables",
            default=False,
            action=argparse.BooleanOptionalAction,
            display_name=self.remove_html_tables_txt,
            help=self.remove_html_tables_hlp,
        )
```
I've yet to test it.


mizaki:
Obviously QT_TRANSLATE_NOOP isn't going to work without Qt...

Easy answer is add the following to comictalker class:
```
    def QT_TRANSLATE_NOOP(self, id, text):
        return text
```

mizaki:
For "external" plugins the easiest option would be to load another translator `<plugin>_talker_translator.load()`.

It would be nice to be able to pull external plugins in to weblate/transifex if possible to consolidate translation.

lordwelch:
> 2. I've cleaned and consolidated the names between the UI files and class names. Also some other cleaning.

At the very least this should be it's own commit if not it's own PR but most of these are irrelevant as long as the loadUI function is used

> 3. A reload of the program is the easiest way to switch languages. Otherwise (afaik) you need to manually re-translate every label.

Qt has a retranslateUi function that gets generated when python files are generated from ui files

> 4. I've included a lang_utils for testing but weblate or transifex will pull the strings for translation in and it appears a GitHub action can be done to generate the .gm files.

qm files are entirely generated and should not be in git. And creating a .pro file and then using pylupdate5 on the pro file would remove the need for the lang_utils module at all

> 5. Some of the lists, such as format/rating/etc. need to be switched to using userData in English so the strings can be translated. (I think this makes sense to do?)

how most translation systems work is that there is a function eg `tr` that marks a string as translatable so that a tool can extract the string to a translation file, note that no translation has happened yet, right now this is purely a marker. When the application runs that string eventually gets sent to an actual translation function which does a simple replacement to the string defined for another language when it is shown to the user. The marking function `tr` doesn't ever need to do anything other than return the string unchanged, most of the time wherever the string is declared is where the translation happens but it does not have to happen there. Typically any comparison means that a string is not translatable. A common practice is also to not have a useable string at all in the source and only have an ID used the real string being listed in the translation file.

> 6. I've put in a test en_GB file (`read style` is changed to `read type`) and a custom message when switching to it. I've also included Japanese (but with no files for CT) to show the builtin translations for the widgets works.

en_GB is fine for a test but under no circumstances am I supporting CT being translated into a separate dialect of english, Qt may have some differences and it's builtin widgets may change but for CT everything will fallback to en_US

> 8. Because this is using the Qt library, no CLI translations. For that gettext appears to be the answer (and switch the GUI over to it too).

It's not uncommon for CLI tools to not have any translations but the GUI tools do

> 9. How to support plugins? Because settngs is used for the strings for CLI and GUI, it presents a problem. For a string to be pulled by pylupdate for translation it needs to be wrapped in `tr()` which requires Qt. Hoping for a bright idea. Best I have is using `QT_TRANSLATE_NOOP` which would look something like:

Well cli options `--comicvine-key` would not change they are not translatable. The help and display name are translatable an as explained in 5. `QT_TRANSLATE_NOOP` is just a marker to say this string is translatable no Qt required, the plugin would provide it's own translations that we would load eg comicvine/i18n/en_GB.qm

and we can't use fstrings for things that need translations!

This is not a trivial project and requires a huge amount of planning and maintenance, even when utilizing weblate or transifex to assist. For some more information/context see [Mozilla's best practices] and [orange3] a python project trying to figure out if they are going to do translations and how. The core developers of orange3 are Slovenian, they created it, they've had it for at least 10 years and it's still only in english! 56% of [Slovenians know english] and their audience is primarily scientific/academic so it's not necessarily a huge jump for them to do so, but english still isn't even an official language of Slovenia.

So unless there are people asking for it (there aren't, this is the first I've seen regarding translation of ComicTagger) I don't want to re-architect the project to be translatable, and I don't know any other languages so if someone opens an issue about a translation all I can say is "sorry, I only know english and I don't have the time or interest to learn an arbitrary language or google translate a better translation". If there are people who want it translated please start a [discussion] or open an [issue] even more so if they would be open to helping translate.

[Mozilla's best practices]: https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#make-strings-properly-localizable
[orange3]: https://github.com/biolab/orange3/issues/5409
[Slovenians know english]: https://en.wikipedia.org/wiki/Slovenia#Languages
[discussion]: https://github.com/comictagger/comictagger/discussions
[issue]: https://github.com/comictagger/comictagger/issues

PR #491, opened 2023-07-17
Use talker pre-calculated hash if available

mizaki:
Adds `ahash` and `phash` to ComicIssue. It may make more sense to use a single field and require some kind of identification. From: https://github.com/comictagger/comictagger/discussions/483
```
What format should the hash field on ComicIssue take? As above in the first message, JSON, list, dict? A dict such as {"ahash": int, "phash": int} might be best?
```
Also, should alt covers be checked or not is the pre-calculated talker hash is present?


lordwelch:
Before adding phash support to IssueIdentifier it needs to be re-worked to handle remote hashes being given to CT instead of downloading the image and hashing it. A remote providing a hash does not impact what images we check, all it does is reduce the bandwidth required to obtain the hash for an image.

IssueIdentifier needs to have clearly separated stages in order to add this, currently it does not and adding this makes it more complicated and if dhash gets added in the future there is no clear path on what would be needed to add it.

PR #500, opened 2023-08-19
Support multiple save styles in GUI

mizaki:
image
Only one read style is allowed for now. It's how to do the ordering for multiple with the overlay order.

CLI supports multiple styles already but I've not touched it due to the argparse question: I've added commandline option for both read and write but also left in the runtime override. How would you like to handle that? Not have those GUI options available on the commandline?

Should the options also be in the settings menu? It may be odd to duplicate them but it also seems odd for them not to be there.

I believe Copy Tags is working but I'm not entirely sure what it's meant to do? Doesn't appear to be anything in the Wiki.


lordwelch:
> Only one read style is allowed for now.

That's fine

> CLI supports multiple styles already but I've not touched it due to the argparse question: I've added commandline option for both read and write but also left in the runtime override. How would you like to handle that? Not have those GUI options available on the commandline?

The internal settings should stay, the type can just be changed. In general if it is something that only affects the GUI it should probably not be a commandline option, I'm sure there are some right now that don't follow that but that's because they should probably affect the commandline as well. `--darkmode` is an obvious exception.

> I believe Copy Tags is working but I'm not entirely sure what it's meant to do? Doesn't appear to be anything in the Wiki.

Copy Tags is to copy from the current load style to the current save style.

All user facing terminology should be save/load/modify tags

I don't really like the table. I think having radio buttons or a multi-select combo box will work better.

Don't add CoMet support in the GUI for this PR, that can be added later.

Looping over the metadata styles and adding extra conditions inside of loops has made the code much more difficult to follow, stick to doing an if else chain. For example in `update_style_tweaks` this is much more readable
```python
        # Enable all items if both metadata styles are selected
        # save_data_styles will always have at least 1 element
        if len(self.save_data_styles) > 1:
            for item in itertools.chain(cix_only, cbi_only):
                enable_widget(item, True)

        elif self.save_data_styles[0] == MetaDataStyle.CIX:
            for item in cix_only:
                enable_widget(item, True)
            for item in cbi_only:
                enable_widget(item, False)

        elif self.save_data_styles[0] == MetaDataStyle.CBI:
            for item in cbi_only:
                enable_widget(item, True)
            for item in cix_only:
                enable_widget(item, False)
```

mizaki:
Thinking on it more, should these options even be on the tagger window? How often are they going to be changed? Default to read/write CIX and move it into settings?

When you say you radio buttons, I can't picture that without them being in a table layout?

Multi-select is doable but that doesn't seem like a good way as it goes against how I would believe most people think a select dropdown box works.

mizaki:
I'll take another look at this when #532 is in but https://gis.stackexchange.com/questions/350148/qcombobox-multiple-selection-pyqt5 seems like an option. The only issue I see is the names running off the end (when closed) if more than a couple are chosen but maybe there isn't a way around that.

PR #503, opened 2023-08-30
PyrateLimter version 2 only for now.

### LINKED ISSUE #502: PyRate latest version breaks Pip install
If you install using ComicTagger using pip it grabs the latest version of PyRate in the setup.cfg file. Running the process results in an error regarding RequestRate not existing as it seems to be removed. To resolve this and fix the install process the version should be locked to a supported version in the setup.cfg or support for the latest version should be added. 

For reference this is the PR with the breaking changes that was merged a few days ago:
https://github.com/vutran1710/PyrateLimiter/pull/111

mizaki:
For now require pyrate-limiter version 2. 
Closes #502

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -41,7 +41,7 @@ install_requires =
     pathvalidate
     pillow>=9.1.0,<10
     pycountry
-    pyrate-limiter
+    pyrate-limiter>=2.6,<3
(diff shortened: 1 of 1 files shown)


lordwelch:
Thanks I hadn't gotten to it yet

PR #518, opened 2023-09-06
Protofolius issue scheme

lordwelch:
Implement Protofolius's issue scheme


mizaki:
This may or may not be related but I was thinking what if there is an option to use your own regex? Keep the current parsers but people can also choose "regex" and enter their own. `^(?P<title>.*?) #(?P<number>\d+) \((?P<year>\d{4})\)$` for example with `Title #1 (2023)`

protofolius:
I tested out the new number scheme that allows issue numbers to start with a letter (was looking forward to this :) )

The preview option however is doing it right , but once I use a real file the parser trips and is not doing it like it does in the preview .

Added 2 screenshots .
Parser Preview
Parser Realfile

protofolius:
To make it work in CLI , use extra option:

--protofolius-issue-number-scheme

PR #529, opened 2023-09-10
Update series and issue rows after calling for more information.

### LINKED ISSUE #512: Issue and series window do not update with new information
When a full series or issue call is made via `current_item_changed`, only a limited number of fields are updated.

Series updates image and desc.
Issues updates issue number, image and desc.

Also, issue only displays month. Year is over written by month looking at it.

mizaki:
I also changed the cursor to "busy" and added a `try` as there are situations where with a mix of using the cache and the API a talker error would cause a crash.

Closes #512

### DIFF
--- a/comictaggerlib/issueselectionwindow.py
+++ b/comictaggerlib/issueselectionwindow.py
@@ -160,31 +160,10 @@ class IssueSelectionWindow(QtWidgets.QDialog):
 
         for row, issue in enumerate(self.issue_list.values()):
             self.twList.insertRow(row)
+            for i in range(3):
+                self.twList.setItem(row, i, QtWidgets.QTableWidgetItem())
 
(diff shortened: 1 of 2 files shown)


lordwelch:
This looks good, but I want the row updates to be separated out into it's own function and call it in the initial update and current_item_changed
E.G. `def update_row(row: int, issue: GenericMetadata) -> None:` and
`def update_row(row: int, series: ComicSeries) -> None:`

mizaki:
I presume this is what you were after https://github.com/comictagger/comictagger/pull/529/commits/366e9cf6e89f7caca3b9860ac0676c7f5f97cd24 ? I also added `title` to the trigger for updating the issue. With that the option in metron_talker wouldn't be needed for the alt images.

PR #532, opened 2023-09-15
Add plugin support for metadata

lordwelch:


comictaggerlib/issueidentifier.py
```diff
@@ -229,10 +229,10 @@ def get_search_keys(self) -> SearchKeys:
...
         # see if the archive has any useful meta data for searching with
         try:
-            if ca.has_cix():
-                internal_metadata = ca.read_cix()
-            else:
-                internal_metadata = ca.read_cbi()
+            for style in metadata_styles:
+                internal_metadata = ca.read_metadata(style)
+                if not internal_metadata.is_empty:
+                    break
```
mizaki, line 235: It doesn't seem like there is a condition where `only_use_additional_meta_data` is ever `False` so this will be reached?

It also breaks on the first found metadata. Maybe not for this PR but part of the overlay/overwrite etc. work it should check each type if it's there? Leave a TODO?

lordwelch: Yeah, I plan on removing the entire option later during a re-organization of the issue identifier.
I changed ComicArchiver to return an empty GenericMetadata if the archive doesn't have the requested metadata style, that way it stays consistent with the previous code.

comicapi/metadata/comet.py
```diff
@@ -0,0 +1,315 @@
...
+            assign("readingDirection", "rtl")
+
+        if md.year is not None:
+            date_str = f"{md.year:04}"
+            if md.month is not None:
+                date_str += f"-{md.month:02}"
+            assign("date", date_str)
+
+        page = md.get_cover_page_index_list()[0]
+        assign("coverImage", md.pages[page]["filename"])
```
mizaki, line 205: `md.pages` is empty then using the CLI `-t comet` but is filled when using the GUI.

```
/comictagger/comicapi/metadata/comet.py", line 198, in _convert_metadata_to_xml
    assign("coverImage", md.pages[page]["filename"])
IndexError: list index out of range
```

mizaki (commented):
Apart from the comet CLI issue and the tests taking about 12s longer. It appears to be working fine.

comicapi/metadata/comet.py
lordwelch: What did you run to get this to happen?

mizaki: `-v -s -t comet -o -i -f "Batman: The Brave and the Bold #1.zip"`

mizaki: Auto-tag does the same in the GUI. I presume the pages aren't read when auto-tagging?

lordwelch: I don't think you updated the branch
```plain
git fetch https://github.com/lordwelch/comictagger +metadata-plugin:lordwelch/metadata-plugin
git checkout lordwelch/metadata-plugin
```

mizaki: I did it again and the same result. I tried with a 2000AD too. You can't replicated?

lordwelch: I figured it out it's the `--clear-metadata` option, somehow it got set to default to on a while ago when it shouldn't have and I put it back in bb67ab009e69aab9a7b23ba9df0beff4f4c751a5 and I usually only test with the default options

lordwelch: It is fixed now. I'll probably merge this PR in tomorrow

comicapi/utils.py
```diff
@@ -195,6 +195,21 @@ def path_to_short_str(original_path: pathlib.Path, renamed_path: pathlib.Path |
...
 
+def get_page_name_list(files: list[str]) -> list[str]:
+    # get the list file names in the archive, and sort
+    files = cast(list[str], os_sorted(files))
+
+    # make a sub-list of image files
+    page_list = []
+    for name in files:
+        if (
+            os.path.splitext(name)[1].casefold() in [".jpg", ".jpeg", ".png", ".gif", ".webp"]
```
mizaki, line 206: `avif` too?

PR #533, opened 2023-09-16
Add combo box support to talker settings generator

mizaki:


comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -167,6 +167,20 @@ def generate_password_textbox(option: settngs.Setting, layout: QtWidgets.QGridLa
...
 
 
+def generate_combobox(option: settngs.Setting, layout: QtWidgets.QGridLayout) -> QtWidgets.QComboBox:
+    row = layout.rowCount()
+    lbl = QtWidgets.QLabel(option.display_name)
+    lbl.setToolTip(option.help)
+    layout.addWidget(lbl, row, 0)
+    widget = QtWidgets.QComboBox()
+    for choice in option.choices:  # type: ignore
+        widget.addItem(str(choice).capitalize(), str(choice))
```
lordwelch: What happens if these are acronyms eg CR, CBL, CoMet?
Best to leave the case alone so that the plugin can display it as they want

lordwelch (changes requested):

PR #536, opened 2023-09-21
Generate password box for any settings dest name that end in password

mizaki:
I think this might need to be documented somewhere?


comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -336,6 +336,8 @@ def generate_source_option_tabs(
             elif option._guess_type() is str:
                 if option.choices is not None:
                     current_widget = generate_combobox(option, layout_grid)
+                elif option.dest.lower().endswith("password"):
```
lordwelch: always use casefold

PR #538, opened 2023-09-21
Prevent a crash if None is passed to sanitize_title

mizaki:
Can't rely on talkers to not pass a `None` as a title (came across this for alt titles).


lordwelch:
I'm not going to merge this. Developers will need to follow the API and pass a string. If I have to do it for rarfile, tarfile(in the future) and others that crash when I give it a non-string then so do the developers of plugins for ComicTagger. Everything should be typed, so they will know exactly what to expect from a function and what to give to a function.

PR #547, opened 2023-09-30
FIX: Window sorting

mizaki:
I removed the sorting when creating the update on click. This puts it back.

### DIFF
--- a/comictaggerlib/issueselectionwindow.py
+++ b/comictaggerlib/issueselectionwindow.py
@@ -160,11 +160,9 @@ class IssueSelectionWindow(QtWidgets.QDialog):
 
         for row, issue in enumerate(self.issue_list.values()):
             self.twList.insertRow(row)
-            for i in range(3):
-                if i == 0:
-                    self.twList.setItem(row, i, IssueNumberTableWidgetItem())
-                else:
-                    self.twList.setItem(row, i, QtWidgets.QTableWidgetItem())
+            self.twList.setItem(row, 0, IssueNumberTableWidgetItem())
+            self.twList.setItem(row, 1, QtWidgets.QTableWidgetItem())
+            self.twList.setItem(row, 2, QtWidgets.QTableWidgetItem())
 
             self.update_row(row, issue)
 
--- a/comictaggerlib/issueselectionwindow.py
+++ b/comictaggerlib/issueselectionwindow.py
@@ -161,7 +161,10 @@ class IssueSelectionWindow(QtWidgets.QDialog):
         for row, issue in enumerate(self.issue_list.values()):
             self.twList.insertRow(row)
             for i in range(3):
-                self.twList.setItem(row, i, QtWidgets.QTableWidgetItem())
+                if i == 0:
+                    self.twList.setItem(row, i, IssueNumberTableWidgetItem())
+                else:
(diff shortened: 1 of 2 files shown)


comictaggerlib/seriesselectionwindow.py
```diff
@@ -534,18 +534,20 @@ def update_row(self, row: int, series: ComicSeries) -> None:
...
-        item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
-        item.setData(QtCore.Qt.ItemDataRole.DisplayRole, series.count_of_issues)
-        item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
+        if series.count_of_issues is not None:
+            item_text = f"{series.count_of_issues:04}"
+            item = self.twList.item(row, 2)
+            item.setText(item_text)
+            item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
+            item.setData(QtCore.Qt.ItemDataRole.DisplayRole, series.count_of_issues)
+            item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
```
lordwelch: These should always update, otherwise we can end up with old data. I don't believe it will with the current code but we should do it properly.

comictaggerlib/issueselectionwindow.py
```diff
@@ -161,7 +161,10 @@ def perform_query(self) -> None:
         for row, issue in enumerate(self.issue_list.values()):
             self.twList.insertRow(row)
             for i in range(3):
-                self.twList.setItem(row, i, QtWidgets.QTableWidgetItem())
+                if i == 0:
+                    self.twList.setItem(row, i, IssueNumberTableWidgetItem())
+                else:
+                    self.twList.setItem(row, i, QtWidgets.QTableWidgetItem())
```
lordwelch: The loop should be unwound it even makes for shorter code in this case.

comictaggerlib/seriesselectionwindow.py
```diff
@@ -534,18 +534,20 @@ def update_row(self, row: int, series: ComicSeries) -> None:
...
-        item_text = str(series.start_year)
-        item = self.twList.item(row, 1)
-        item.setText(item_text)
-        item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
-        item.setFlags(QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled)
+        if series.start_year is not None:
+            item_text = f"{series.start_year:04}"
+            item = self.twList.item(row, 1)
+            item.setData(QtCore.Qt.ItemDataRole.ToolTipRole, item_text)
+            item.setData(QtCore.Qt.ItemDataRole.DisplayRole, series.start_year)
```
lordwelch: Why the change from setText to setData

lordwelch (changes requested):

comictaggerlib/seriesselectionwindow.py
mizaki: I imagine it was where I was trying to break apart the display text from the sorting. Reverted.

PR #559, opened 2023-10-23
Add file picker to talkers options

mizaki:
If `dest` ends with `path` add a file picker `browse` button next to the text box. Similar to passwords.

### DIFF
--- a/comictaggerlib/ui/talkeruigenerator.py
+++ b/comictaggerlib/ui/talkeruigenerator.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import logging
 from functools import partial
+from pathlib import Path
 from typing import Any, NamedTuple, cast
 
 import settngs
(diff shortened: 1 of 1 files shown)


comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -338,6 +360,8 @@ def generate_source_option_tabs(
                     current_widget = generate_combobox(option, layout_grid)
                 elif option.dest.casefold().endswith("password"):
                     current_widget = generate_password_textbox(option, layout_grid)
+                elif option.dest.casefold().endswith("path"):
+                    current_widget = generate_path_textbox(option, layout_grid)
```
lordwelch: Instead of checking the name lets check the type to see if it is a `pathlib.Path` object

mizaki: By doing?
```
parser.add_setting(
            "--gcd-filepath",
            display_name="SQLite DB",
            type=pathlib.Path,
            help="The path and filename of the GCD SQLite file",
        )
```
Initially I thought you meant by testing the string but didn't see that working with an empty one. (Setting the default to `Path.home()` would get around that, I suppose...)

lordwelch: For the default they would create a `Path` object or they would just leave it as `None`

lordwelch (approved):

PR #568, opened 2023-11-18
Add custom save function to allow saving of pathlib paths

### LINKED ISSUE #567: Save settings with Path - not serializable
JSON Encoder(/Decoder?) required.

```
TypeError: Object of type PosixPath is not JSON serializable
2023-11-14T22:53:17 | settngs | ERROR | Failed to save config file: .config/ComicTagger/settings.json
  File "/workspace/comictagger/venv/lib/python3.10/site-packages/settngs/__init__.py", line 533, in save_file
    json_str = json.dumps(file_options, indent=2)
  File "/usr/lib64/python3.10/json/__init__.py", line 238, in dumps
    **kw).encode(obj)
  File "/usr/lib64/python3.10/json/encoder.py", line 201, in encode
    chunks = list(chunks)
  File "/usr/lib64/python3.10/json/encoder.py", line 431, in _iterencode
    yield from _iterencode_dict(o, _current_indent_level)
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 405, in _iterencode_dict
    yield from chunks
  File "/usr/lib64/python3.10/json/encoder.py", line 438, in _iterencode
    o = _default(o)
  File "/usr/lib64/python3.10/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type PosixPath is not JSON serializable
```

mizaki:
This seems to be the simplest version. Closes #567

Should the talker be expected to convert over as part of `parse_settings`?

The other option would be to save in a format such as `{"type": "pathlib.Path", "value": "/my_home/"}` and then have a JSON decoder handle the conversion back. That could require using `import` based on the saved type which seems bad.

### DIFF
--- a/comictaggerlib/ctsettings/__init__.py
+++ b/comictaggerlib/ctsettings/__init__.py
@@ -1,5 +1,12 @@
 from __future__ import annotations
 
+import json
+import logging
+import pathlib
+from typing import Any
+
+import settngs
+
 from comictaggerlib.ctsettings.commandline import (
     initial_commandline_parser,
     register_commandline_settings,
@@ -10,6 +17,8 @@ from comictaggerlib.ctsettings.plugin import group_for_plugin, register_plugin_s
 from comictaggerlib.ctsettings.settngs_namespace import settngs_namespace as ct_ns
 from comictaggerlib.ctsettings.types import ComicTaggerPaths
 from comictalker import ComicTalker
+
+logger = logging.getLogger(__name__)
 
 talkers: dict[str, ComicTalker] = {}
 
@@ -25,3 +34,79 @@ __all__ = [
     "ct_ns",
     "group_for_plugin",
 ]
+
+
+class SettingsEncoder(json.JSONEncoder):
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, pathlib.Path):
+            return str(obj)
+
+        # Let the base class default method raise the TypeError
+        return json.JSONEncoder.default(self, obj)
+
+
+def validate_types(config: settngs.Config[settngs.Values]) -> settngs.Config[settngs.Values]:
+    # Go through each setting
+    for group in config.definitions.values():
+        for setting in group.v.values():
+            # Get the value and if it is the default
+            value, default = settngs.get_option(config.values, setting)
+            if not default:
+                if setting.type is not None:
+                    # If it is not the default and the type attribute is not None
+                    # use it to convert the loaded string into the expected value
+                    if isinstance(value, str):
+                        config.values[setting.group][setting.dest] = setting.type(value)
+    return config
+
+
+def parse_config(
+    manager: settngs.Manager,
+    config_path: pathlib.Path,
+    args: list[str] | None = None,
+) -> tuple[settngs.Config[settngs.Values], bool]:
+    """
+    Function to parse options from a json file and passes the resulting Config object to parse_cmdline.
+
+    Args:
+        manager: settngs Manager object
+        config_path: A `pathlib.Path` object
+        args: Passed to argparse.ArgumentParser.parse_args
+    """
+    file_options, success = settngs.parse_file(manager.definitions, config_path)
+    file_options = validate_types(file_options)
+    cmdline_options = settngs.parse_cmdline(
+        manager.definitions,
+        manager.description,
+        manager.epilog,
+        args,
+        file_options,
+    )
+
+    final_options = settngs.normalize_config(cmdline_options, file=True, cmdline=True)
+    return final_options, success
+
+
+def save_file(
+    config: settngs.Config[settngs.T],
+    filename: pathlib.Path,
+) -> bool:
+    """
+    Helper function to save options from a json dictionary to a file
+
+    Args:
+        config: The options to save to a json dictionary
+        filename: A pathlib.Path object to save the json dictionary to
+    """
+    file_options = settngs.clean_config(config, file=True)
+    try:
+        if not filename.exists():
+            filename.parent.mkdir(exist_ok=True, parents=True)
+            filename.touch()
+
+        json_str = json.dumps(file_options, cls=SettingsEncoder, indent=2)
+        filename.write_text(json_str + "\n", encoding="utf-8")
+    except Exception:
+        logger.exception("Failed to save config file: %s", filename)
+        return False
+    return True
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -150,8 +150,8 @@ class App:
         ctsettings.register_plugin_settings(self.manager)
 
     def parse_settings(self, config_paths: ctsettings.ComicTaggerPaths, *args: str) -> settngs.Config[ct_ns]:
-        cfg, self.config_load_success = self.manager.parse_config(
-            config_paths.user_config_dir / "settings.json", list(args) or None
+        cfg, self.config_load_success = ctsettings.parse_config(
+            self.manager, config_paths.user_config_dir / "settings.json", list(args) or None
         )
         config = cast(settngs.Config[ct_ns], self.manager.get_namespace(cfg, file=True, cmdline=True))
         config[0].Runtime_Options__config = config_paths
@@ -206,8 +206,8 @@ class App:
             if self.config_load_success:
                 settings_path = self.config[0].Runtime_Options__config.user_config_dir / "settings.json"
                 if self.config_load_success:
-                    self.manager.save_file(self.config[0], settings_path)
-                print("Key set")  # noqa: T201
+                    ctsettings.save_file(self.config, settings_path)
+                print("Settings saved")  # noqa: T201
                 return
 
         if not self.config_load_success:
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -39,7 +39,7 @@ from comicapi.comicinfoxml import ComicInfoXml
 from comicapi.filenameparser import FileNameParser
 from comicapi.genericmetadata import GenericMetadata
 from comicapi.issuestring import IssueString
-from comictaggerlib import ctversion
+from comictaggerlib import ctsettings, ctversion
 from comictaggerlib.applicationlogwindow import ApplicationLogWindow, QTextEditLogger
 from comictaggerlib.autotagmatchwindow import AutoTagMatchWindow
 from comictaggerlib.autotagprogresswindow import AutoTagProgressWindow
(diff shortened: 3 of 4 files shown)


comictaggerlib/ctsettings/__init__.py
```diff
@@ -25,3 +34,37 @@
...
 ]
+
+
+class SettingsEncoder(json.JSONEncoder):
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, pathlib.Path):
+            return str(obj)
+
+        # Otherwise raise standard error
+        raise TypeError(f"Object of type {obj.__class__.__name__} is not JSON serializable")
```
lordwelch: I don't believe this will work I believe you need use the parent `default` method otherwise only Pathlib.Path objects will get converted

comictaggerlib/main.py
```diff
@@ -206,8 +206,8 @@ def main(self) -> None:
             if self.config_load_success:
                 settings_path = self.config[0].Runtime_Options_config.user_config_dir / "settings.json"
                 if self.config_load_success:
-                    self.manager.save_file(self.config[0], settings_path)
-                print("Key set")  # noqa: T201
+                    save_file(self.config, settings_path)
```
lordwelch: This needs to be renamed either use the full import `ctsettings.save_file` or rename this to save_settings or a similar name

comictaggerlib/taggerwindow.py
```diff
@@ -29,6 +29,7 @@
 from typing import Any, Callable
 from urllib.parse import urlparse
 
+import ctsettings
```
lordwelch: This is the wrong import. I cannot load the GUI
```plain
2023-11-17T19:25:43 | comictagger | ERROR | Qt unavailable
  File "/Users/timmy/build/source/comictagger/comictaggerlib/gui.py", line 64, in <module>
    from comictaggerlib.taggerwindow import TaggerWindow
  File "/Users/timmy/build/source/comictagger/comictaggerlib/taggerwindow.py", line 32, in <module>
    import ctsettings
ModuleNotFoundError: No module named 'ctsettings'
2023-11-17T19:25:43 | comictagger | WARNING | PyQt5 is not available. ComicTagger is limited to command-line mode.
2023-11-17T19:25:43 | comictaggerlib.cli | ERROR | You must specify at least one filename.  Use the -h option for more info
```

comictaggerlib/ui/talkeruigenerator.py
```diff
@@ -259,6 +260,13 @@ def settings_to_talker_form(sources: Sources, config: settngs.Config[ct_ns]) ->
             try:
                 if isinstance(value, str) and value and isinstance(widget, QtWidgets.QLineEdit) and not default:
                     widget.setText(value)
+                if (
+                    isinstance(value, pathlib.Path)
+                    and value
+                    and isinstance(widget, QtWidgets.QLineEdit)
+                    and not default
+                ):
+                    widget.setText(str(value))
```
lordwelch: Put this in a separate PR and add a comment explaining what is happening

lordwelch (changes requested):
Se we also need a decoder. We can use the same method that we use for the ui generator, take the `type` attribute from the setting definition if it exists and run the string through it

comictaggerlib/ctsettings/__init__.py
mizaki: Looking at the output bool, int etc. all seem to have been converted. As I understand it, it will try the internal supported conversions and then use the custom class. `default` only raises a type error in the base. I will change the error raise as suggest here though: https://docs.python.org/3/library/json.html#json.JSONEncoder.default

mizaki:
> Se we also need a decoder. We can use the same method that we use for the ui generator, take the `type` attribute from the setting definition if it exists and run the string through it

I forgot to include the `parse_settings` I was using `settings["gcd_filepath"] = pathlib.Path(settings["gcd_filepath"])`. I was thinking that it might be possible the module won't have been loaded and so it might error? Having the talker do it via `parse_settings` would get around it? I'll try it and see.

mizaki:
> > Se we also need a decoder. We can use the same method that we use for the ui generator, take the `type` attribute from the setting definition if it exists and run the string through it
> 
> I forgot to include the `parse_settings` I was using `settings["gcd_filepath"] = pathlib.Path(settings["gcd_filepath"])`. I was thinking that it might be possible the module won't have been loaded and so it might error? Having the talker do it via `parse_settings` would get around it? I'll try it and see.

Looking at it, unless I've missed it (quite possible) a decoder is not all that viable. Because the file from `main` is parsed through the `manager` using `parse_config` (which calls the base `parse_config` which calls `parse_file`) so a custom `parse_config` that calls a custom `parse_file` that calls `json.loads` with a `json.decoder` is required however, the decoder does not get the `definitions`.

`validate_talker_settings` might be a sensible place to do the type conversion but that calls the talkers `parse_settings` so we could have `parse_settings` do any required conversion. As I've done currently with GCD.

This is with `pathlib` in mind which does not require any decoding because it can be encoded to a string. If there were a more complex type that requires JSON decoding then the custom `parse_settings` is required again. What other types can we reasonably expect to have though?

As doing it in the decoder seems out, which of these would you prefer:
1. Have the talker's `parse_settings` handle any type conversion
2. Have `plugin.py` `validate_talker_settings` check and make any type conversion
3. You have a better idea

Hopefully all that is clear enough and makes sense.

lordwelch:
> Looking at it, unless I've missed it (quite possible) a decoder is not all that viable. Because the file from main is parsed through the manager using parse_config (which calls the base parse_config which calls parse_file) so a custom parse_config that calls a custom parse_file that calls json.loads with a json.decoder is required however, the decoder does not get the definitions.

I made the parse_config and parse_file functions purposefully incredibly basic convenience functions.
The parse_config function has this documentation
>        Convenience function to parse options from a json file and passes the resulting Config object to parse_cmdline.
>        This is purely a convenience function.
>        If _anything_ more advanced is desired this should be handled by the application.

The entire function is:
```python
    file_options, success = parse_file(definitions, config_path)
    cmdline_options = parse_cmdline(
        definitions, description, epilog, args, file_options,
    )

    final_options = normalize_config(cmdline_options, file=True, cmdline=True)
    return final_options, success
```

There are already supported methods for doing type conversions from a single string to whatever type they expect https://docs.python.org/3/library/argparse.html#quick-links-for-add-argument

Which is re-used by CT right here:
https://github.com/comictagger/comictagger/blob/57873136b64abc74c5dd0c551739334cb90d26ee/comictaggerlib/ui/talkeruigenerator.py#L292-L295

What is actually more problematic is converting from an arbitrary object to a string or other json compatible representation.
Doing this for a pathlib.Path object is easier because it's in the standard library and is a glorified string.

> the decoder does not get the definitions

This is mostly true, the json.loads and json.dumps methods basically just fill out the constructor for the json decoder/encoder and then call decode/encode https://github.com/python/cpython/blob/3.12/Lib/json/__init__.py and then they make it easy by passing an kwargs through to the cls argument which means we can do something like this:
```python
class CTDecoder(json.JSONDecoder):
    def __init__(*args, definitions: settngs.Definitions, **kwargs)
        super().__init__(*args, **kwargs)
        self.definitions = definitions
    def decode(s) -> Any:
        ...
        
loaded_options = json.loads(json_string, cls=CTDecoder, definitions=definitions)  # definitions gets passed through to CTDecoder
```

mizaki:
Is this more like it?

Should I try and generalise the encoding? Use `__dict__` say or manually craft as each type is required?

comictaggerlib/ctsettings/__init__.py
```diff
@@ -25,3 +34,115 @@
     "ct_ns",
     "group_for_plugin",
 ]
+
+
+class SettingsEncoder(json.JSONEncoder):
+    def default(self, obj: Any) -> Any:
+        if isinstance(obj, pathlib.Path):
+            return {"type": "pathlib.Path", "value": str(obj)}
```
lordwelch:
```suggestion
            return str(obj)
```

comictaggerlib/ctsettings/__init__.py
```diff
@@ -25,3 +34,115 @@
...
+    return settngs.normalize_config(settngs.Config(options, definitions), file=True), success
+
+
+def parse_config(
+    definitions: settngs.Definitions,
+    description: str,
+    epilog: str,
+    config_path: pathlib.Path,
+    args: list[str] | None = None,
+) -> tuple[settngs.Config[settngs.Values], bool]:
```
lordwelch: This should take in the entire manager, there is no reason to list every individual argument here

comictaggerlib/main.py
```diff
@@ -150,8 +150,8 @@ def register_settings(self) -> None:
         ctsettings.register_plugin_settings(self.manager)
 
     def parse_settings(self, config_paths: ctsettings.ComicTaggerPaths, *args: str) -> settngs.Config[ct_ns]:
-        cfg, self.config_load_success = self.manager.parse_config(
-            config_paths.user_config_dir / "settings.json", list(args) or None
+        cfg, self.config_load_success = ctsettings.parse_config(
+            self.manager.definitions, "", "", config_paths.user_config_dir / "settings.json", list(args) or None
```
lordwelch: The manager should be passed into parse_config

comictaggerlib/ctsettings/__init__.py
```diff
@@ -25,3 +34,115 @@
...
+            # The source name is required, use the object with the full file
+            if o.startswith("Source ") and isinstance(v, dict):
+                for key, value in v.items():
+                    if isinstance(value, dict) and value.get("type"):
+                        typ = self.definitions[o].v[key].type
+                        if typ:
+                            # Should check types match?
+                            obj[o][key] = typ(value["value"])
+
+        return obj
```
lordwelch: After looking at it the JSON api doesn't provide an option that gives enough context. Instead it can just be processed after it's loaded from json.
eg
```python
def validate_types(options: settngs.Config[settngs.Values], definitions: settngs.Definitions) -> settngs.Config[settngs.Values]:
    # Go through each setting
    for group in definitions.values():
        for setting in group.v.values():
            # Get the value and if it is the default
            value, default = settngs.get_option(options.values, setting)
            if not default:
                if setting.type is not None:
                    # If it is not the default and the type attribute is not None
                    # use it to convert the loaded string into the expected value
                    if isinstance(value, str):
                        options.values[setting.group][setting.dest] = setting.type(value)
    return options
    ```

**Review by lordwelch — CHANGES_REQUESTED** [review]

**Comment by mizaki** [comment]
I did a squish too.

File: comictaggerlib/ctsettings/__init__.py [hunk 9]
```diff
@@ -25,3 +34,106 @@
...
+            else:  # pragma: no cover
+                raise Exception("Loaded file is not a JSON Dictionary")
+        except Exception:  # pragma: no cover
+            logger.exception("Failed to load config file: %s", filename)
+            success = False
+    else:
+        logger.info("No config file found")
+        success = True
+
+    return settngs.normalize_config(validate_types(settngs.Config(options, definitions)), file=True), success
```
lordwelch: Is there a reason to not use the function from settngs? validate_types does not need to happen in this function.

mizaki: Nope. Removed and squished.

lordwelch (approved):

PR #570, opened 2023-11-18
Add support for pathlib.Path in talker settings with file picker

mizaki:


lordwelch:
I have a slightly simpler way to fix this and I have some other changes I'm putting in as well.

PR #573, opened 2023-11-30
Add CT verion check against talker requirements

mizaki:
I picked `>=1.6.0a7` as the default because that is when the lists were changed to sets etc.

### DIFF
--- a/comictalker/__init__.py
+++ b/comictalker/__init__.py
@@ -4,7 +4,7 @@ import logging
 import pathlib
 import sys
 
-from packaging.specifiers import InvalidSpecifier, SpecifierSet
+from packaging.version import InvalidVersion, parse
 
 if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
@@ -33,14 +33,14 @@ def get_talkers(version: str, cache: pathlib.Path) -> dict[str, ComicTalker]:
                 logger.error("Talker ID must be the same as the entry point name")
                 continue
             try:
-                if version in SpecifierSet(obj.ct_req_spec, prereleases=True):
+                if parse(version) >= parse(obj.comictagger_min_ver):
                     talkers[talker.name] = obj
                 else:
                     logger.error(
-                        f"CT required version not met for talker: {talker.name} with specifier: {obj.ct_req_spec}"
+                        f"CT minimum required version not met for talker: {talker.name} with version: {obj.comictagger_min_ver}"
                     )
-            except InvalidSpecifier:
-                logger.error(f"Invalid specifier for talker: {talker.name} - specifier: {obj.ct_req_spec}")
+            except InvalidVersion:
+                logger.error(f"Invalid version number for talker: {talker.name} - version: {obj.comictagger_min_ver}")
 
         except Exception:
             logger.exception("Failed to load talker: %s", talker.name)
--- a/comictalker/__init__.py
+++ b/comictalker/__init__.py
@@ -4,6 +4,8 @@ import logging
 import pathlib
 import sys
 
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
+
 if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
 else:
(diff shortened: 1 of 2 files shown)


mizaki:
Should `ct_req_spec` be an empty string in class `ComicTalker`?

Should the GUI notify if a talker fails to load?

comictalker/comictalker.py
```diff
@@ -107,6 +107,7 @@ class ComicTalker:
 
     name: str = "Example"
     id: str = "example"
+    ct_req_spec: str = ">=1.6.0a7"  # The ComicTagger version required by the talker using PyPA version specifiers
```
lordwelch: ct should be spelled out

comictalker/__init__.py
```diff
@@ -30,7 +32,15 @@ def get_talkers(version: str, cache: pathlib.Path) -> dict[str, ComicTalker]:
...
-            talkers[talker.name] = obj
+            try:
+                if version in SpecifierSet(obj.ct_req_spec, prereleases=True):
+                    talkers[talker.name] = obj
+                else:
+                    logger.error(
+                        f"CT required version not met for talker: {talker.name} with specifier: {obj.ct_req_spec}"
+                    )
+            except InvalidSpecifier:
+                logger.error(f"Invalid specifier for talker: {talker.name} - specifier: {obj.ct_req_spec}")
```
lordwelch: How about we just make it minimum version instead of using the full specification

mizaki: My thinking was in case there are breaking changes a max version could be set.

lordwelch: My rationale is that if there is a breaking change they will have to make a code change to exclude it, and instead they should make the needed code changes

comictalker/__init__.py
```diff
@@ -30,7 +32,15 @@ def get_talkers(version: str, cache: pathlib.Path) -> dict[str, ComicTalker]:
...
-            talkers[talker.name] = obj
+            try:
+                if parse(version) >= parse(obj.comictagger_min_ver):
+                    talkers[talker.name] = obj
+                else:
+                    logger.error(
+                        f"CT minimum required version not met for talker: {talker.name} with version: {obj.comictagger_min_ver}"
+                    )
+            except InvalidVersion:
+                logger.error(f"Invalid version number for talker: {talker.name} - version: {obj.comictagger_min_ver}")
```
lordwelch: I think this should be a warning instead and we should still try to use the plugin

comictalker/__init__.py
```diff
@@ -30,7 +32,15 @@ def get_talkers(version: str, cache: pathlib.Path) -> dict[str, ComicTalker]:
             if obj.id != talker.name:
                 logger.error("Talker ID must be the same as the entry point name")
                 continue
-            talkers[talker.name] = obj
+            try:
+                if parse(version) >= parse(obj.comictagger_min_ver):
+                    talkers[talker.name] = obj
+                else:
+                    logger.error(
+                        f"CT minimum required version not met for talker: {talker.name} with version: {obj.comictagger_min_ver}"
```
lordwelch: CT should be spelled out. To me this reads that the version is the talker version and not ComicTagger's version

lordwelch (commented):
We should list the plugins that are disabled and why eg execution failed, wrong version but doesn't need to be implemented in this PR

lordwelch:
I don't like pulling in the packaging dependency when we just need to parse a basic version
Something like this should fit our needs
```python
import pytest
from typing import Tuple

def parse_version(version_str: str) -> Tuple[int, int, int]:
    # Split the version string into segments and convert them to integers
    version_parts = version_str.split('.')
    parts = [int(seg) if seg.isdigit() else 0 for seg in version_parts[:3]]
    
    # Ensure exactly three elements in the resulting list
    parts.extend([0] * (3 - len(parts)))

    # Return a tuple with the first three elements from the parsed segments
    return (parts[0], parts[1], parts[2])

@pytest.mark.parametrize(
    "version_str, expected",
    [
        ("1.2.3", (1, 2, 3)),       # Test case with a standard version string
        ("0.0.1", (0, 0, 1)),       # Test case with segments containing only zeros except for the last segment
        ("10.5", (10, 5, 0)),       # Test case with only two segments
        ("2", (2, 0, 0)),           # Test case with only one segment
        ("3.7.9.2", (3, 7, 9)),     # Test case with more than three segments
        ("", (0, 0, 0))             # Test case for an empty string
    ],
)
def test_parse_version(version_str, expected):
    # Check if the output of parse_version matches the expected value for each test case
    assert parse_version(version_str) == expected
```

Tuples can then be compared with <,> easily

mizaki:
> I don't like pulling in the packaging dependency when we just need to parse a basic version Something like this should fit our needs

Given the current importance of the 'a<number>' parsing that out to a 4 part number seems sensible?

lordwelch:
You're right, I forgot about the alpha releases, those need to sort before the normal release. Then we'll keep the packaging dependency.

comictalker/__init__.py
```diff
@@ -30,7 +32,20 @@ def get_talkers(version: str, cache: pathlib.Path) -> dict[str, ComicTalker]:
             if obj.id != talker.name:
                 logger.error("Talker ID must be the same as the entry point name")
                 continue
-            talkers[talker.name] = obj
+            try:
+                if parse(version) >= parse(obj.comictagger_min_ver):
```
lordwelch: Move this `parse(version)` out of the loop, we don't need to parse this again for every plugin

lordwelch (changes requested):

lordwelch (approved):

PR #577, opened 2023-12-16
Fix crash when using shortcut to set page type

DrMcCoy:
QListWidget has no rowCount() method, it has count() instead.


lordwelch:
Thanks, I apparently missed that when I fixed the last crash

PR #594, opened 2024-01-27
Add support for saving multiple metadata styles in the GUI

mizaki:
It may make more sense to put the `CheckableComboBox` in its own file in the `ui` dir?

I gave up with the style/theming and it seems a lot of hassle https://forum.qt.io/topic/113522/styling-qcombobox-s-lineedit

I presumed there is no point in allowing no write styles, so one is always required and the tooltip saying as much is displayed if it's tried. I thought this is probably the best place rather than say, checking on save etc.

A couple of places I changed reading of the current archive metadata from the write style to the read style. I can only think the idea was to keep what was already there? But if a read style is picked, then to my mind that should be used.

I'll do the read styles in another PR. With that, it may make sense to add another tab (or section) for the read and write styles in the GUI options.

### DIFF
--- a/comictaggerlib/autotagmatchwindow.py
+++ b/comictaggerlib/autotagmatchwindow.py
@@ -251,6 +251,11 @@ class AutoTagMatchWindow(QtWidgets.QDialog):
             success = ca.write_metadata(md, style)
             QtWidgets.QApplication.restoreOverrideCursor()
             if not success:
-                QtWidgets.QMessageBox.warning(self, "Write Error", "Saving the tags to the archive seemed to fail!")
+                QtWidgets.QMessageBox.warning(
+                    self,
+                    "Write Error",
+                    f"Saving {metadata_styles[style].name()} the tags to the archive seemed to fail!",
+                )
+                break
 
         ca.load_cache(list(metadata_styles))
--- a/comictaggerlib/autotagmatchwindow.py
+++ b/comictaggerlib/autotagmatchwindow.py
@@ -39,7 +39,7 @@ class AutoTagMatchWindow(QtWidgets.QDialog):
         self,
         parent: QtWidgets.QWidget,
         match_set_list: list[Result],
-        style: str,
+        styles: list[str],
         fetch_func: Callable[[IssueResult], GenericMetadata],
         config: ct_ns,
         talker: ComicTalker,
@@ -81,7 +81,7 @@ class AutoTagMatchWindow(QtWidgets.QDialog):
         self.buttonBox.button(QtWidgets.QDialogButtonBox.StandardButton.Ok).setText("Accept and Write Tags")
 
         self.match_set_list = match_set_list
-        self._style = style
+        self._styles = styles
         self.fetch_func = fetch_func
 
         self.current_match_set_idx = 0
@@ -230,7 +230,7 @@ class AutoTagMatchWindow(QtWidgets.QDialog):
         match = self.current_match()
         ca = ComicArchive(self.current_match_set.original_path)
 
-        md = ca.read_metadata(self._style)
+        md = ca.read_metadata(self.config.internal__load_data_style)
         if md.is_empty:
             md = ca.metadata_from_filename(
                 self.config.Filename_Parsing__complicated_parser,
@@ -247,10 +247,10 @@ class AutoTagMatchWindow(QtWidgets.QDialog):
 
         QtWidgets.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
         md.overlay(ct_md)
-        success = ca.write_metadata(md, self._style)
+        for style in self._styles:
+            success = ca.write_metadata(md, style)
+            QtWidgets.QApplication.restoreOverrideCursor()
+            if not success:
+                QtWidgets.QMessageBox.warning(self, "Write Error", "Saving the tags to the archive seemed to fail!")
+
         ca.load_cache(list(metadata_styles))
-
-        QtWidgets.QApplication.restoreOverrideCursor()
-
-        if not success:
-            QtWidgets.QMessageBox.warning(self, "Write Error", "Saving the tags to the archive seemed to fail!")
--- a/comictaggerlib/pagelisteditor.py
+++ b/comictaggerlib/pagelisteditor.py
@@ -115,7 +115,7 @@ class PageListEditor(QtWidgets.QWidget):
 
         self.comic_archive: ComicArchive | None = None
         self.pages_list: list[ImageMetadata] = []
-        self.data_style = ""
+        self.data_styles: list[str] = []
 
     def reset_page(self) -> None:
         self.pageWidget.clear()
@@ -326,7 +326,7 @@ class PageListEditor(QtWidgets.QWidget):
         self.comic_archive = comic_archive
         self.pages_list = pages_list
         if pages_list:
-            self.set_metadata_style(self.data_style)
+            self.set_metadata_style(self.data_styles)
         else:
             self.cbPageType.setEnabled(False)
             self.chkDoublePage.setEnabled(False)
@@ -370,11 +370,16 @@ class PageListEditor(QtWidgets.QWidget):
             self.first_front_page = self.get_first_front_cover()
             self.firstFrontCoverChanged.emit(self.first_front_page)
 
-    def set_metadata_style(self, data_style: str) -> None:
+    def set_metadata_style(self, data_styles: list[str]) -> None:
         # depending on the current data style, certain fields are disabled
-        if data_style:
-            self.metadata_style = data_style
+        if data_styles:
+            styles = [metadata_styles[style] for style in data_styles]
+            enabled_widgets = []
+            for style in styles:
+                for attr in style.supported_attributes:
+                    enabled_widgets.append(attr)
 
-            enabled_widgets = metadata_styles[data_style].supported_attributes
+            self.data_styles = data_styles
+
             for metadata, widget in self.md_attributes.items():
                 enable_widget(widget, metadata in enabled_widgets)
--- a/comictaggerlib/pagelisteditor.py
+++ b/comictaggerlib/pagelisteditor.py
@@ -374,10 +374,9 @@ class PageListEditor(QtWidgets.QWidget):
         # depending on the current data style, certain fields are disabled
         if data_styles:
             styles = [metadata_styles[style] for style in data_styles]
-            enabled_widgets = []
+            enabled_widgets = set()
             for style in styles:
-                for attr in style.supported_attributes:
-                    enabled_widgets.append(attr)
+                enabled_widgets.update(style.supported_attributes)
 
             self.data_styles = data_styles
 
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -61,6 +61,7 @@ from comictaggerlib.resulttypes import Action, IssueResult, MatchStatus, OnlineM
 from comictaggerlib.seriesselectionwindow import SeriesSelectionWindow
 from comictaggerlib.settingswindow import SettingsWindow
 from comictaggerlib.ui import ui_path
+from comictaggerlib.ui.customwidgets import CheckableComboBox
 from comictaggerlib.ui.qtutils import center_window_on_parent, enable_widget, reduce_widget_font_size
 from comictaggerlib.versionchecker import VersionChecker
 from comictalker.comictalker import ComicTalker, TalkerError
@@ -215,15 +216,29 @@ class TaggerWindow(QtWidgets.QMainWindow):
 
         if config[0].Runtime_Options__type and isinstance(config[0].Runtime_Options__type[0], str):
             # respect the command line option tag type
-            config[0].internal__save_data_style = config[0].Runtime_Options__type[0]
+            config[0].internal__save_data_style = config[0].Runtime_Options__type
             config[0].internal__load_data_style = config[0].Runtime_Options__type[0]
 
-        if config[0].internal__save_data_style not in metadata_styles:
-            config[0].internal__save_data_style = list(metadata_styles.keys())[0]
+        for style in config[0].internal__save_data_style:
+            if style not in metadata_styles:
+                config[0].internal__save_data_style.remove(style)
         if config[0].internal__load_data_style not in metadata_styles:
             config[0].internal__load_data_style = list(metadata_styles.keys())[0]
-        self.save_data_style = config[0].internal__save_data_style
-        self.load_data_style = config[0].internal__load_data_style
+        self.save_data_styles: list[str] = config[0].internal__save_data_style
+        self.load_data_style: str = config[0].internal__load_data_style
+
+        # Add multiselect combobox
+        self.cbSaveDataStyle = CheckableComboBox()
+        self.cbSaveDataStyle.setToolTip("At least 1 save style is required")
+        # Add normal combobox for read style (TODO support multiple read styles)
+        self.cbLoadDataStyle = QtWidgets.QComboBox()
+
+        # Need to set minimum or source_style_formLayout will resize larger than 230px which will affect the
+        # file info box and cover image width underneath
+        self.cbLoadDataStyle.setMinimumWidth(100)
+        self.cbSaveDataStyle.setMinimumWidth(100)
+        self.source_style_formLayout.addRow("Read Style", self.cbLoadDataStyle)
+        self.source_style_formLayout.addRow("Modify Styles", self.cbSaveDataStyle)
 
         self.setAcceptDrops(True)
         self.view_tag_actions, self.remove_tag_actions = self.tag_actions()
@@ -273,7 +288,7 @@ class TaggerWindow(QtWidgets.QMainWindow):
 
         # hook up the callbacks
         self.cbLoadDataStyle.currentIndexChanged.connect(self.set_load_data_style)
-        self.cbSaveDataStyle.currentIndexChanged.connect(self.set_save_data_style)
+        self.cbSaveDataStyle.itemChecked.connect(self.set_save_data_style)
         self.cbx_sources.currentIndexChanged.connect(self.set_source)
         self.btnEditCredit.clicked.connect(self.edit_credit)
         self.btnAddCredit.clicked.connect(self.add_credit)
@@ -438,7 +453,7 @@ class TaggerWindow(QtWidgets.QMainWindow):
         self.actionCopyTags.triggered.connect(self.copy_tags)
 
         self.actionRemoveAuto.setShortcut("Ctrl+D")
-        self.actionRemoveAuto.setStatusTip("Remove currently selected modify tag style from the archive")
+        self.actionRemoveAuto.setStatusTip("Remove currently selected modify tag style(s) from the archive")
         self.actionRemoveAuto.triggered.connect(self.remove_auto)
 
         self.actionRepackage.setShortcut("Ctrl+E")
@@ -1146,7 +1161,7 @@ class TaggerWindow(QtWidgets.QMainWindow):
             reply = QtWidgets.QMessageBox.question(
                 self,
                 "Save Tags",
-                f"Are you sure you wish to save {metadata_styles[self.save_data_style].name()} tags to this archive?",
+                f"Are you sure you wish to save {', '.join([metadata_styles[style].name() for style in self.save_data_styles])} tags to this archive?",
                 QtWidgets.QMessageBox.StandardButton.Yes,
                 QtWidgets.QMessageBox.StandardButton.No,
             )
@@ -1155,12 +1170,22 @@ class TaggerWindow(QtWidgets.QMainWindow):
                 QtWidgets.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
                 self.form_to_metadata()
 
-                success = self.comic_archive.write_metadata(self.metadata, self.save_data_style)
+                fail_list = []
+                # Save each style
+                for style in self.save_data_styles:
+                    success = self.comic_archive.write_metadata(self.metadata, style)
+                    if not success:
+                        fail_list.append(style)
+
                 self.comic_archive.load_cache(list(metadata_styles))
                 QtWidgets.QApplication.restoreOverrideCursor()
 
-                if not success:
-                    QtWidgets.QMessageBox.warning(self, "Save failed", "The tag save operation seemed to fail!")
+                if fail_list:
+                    QtWidgets.QMessageBox.warning(
+                        self,
+                        "Save failed",
+                        f"The tag save operation seemed to fail for:{', '.join([metadata_styles[style].name() for style in fail_list])}",
+                    )
                 else:
                     self.clear_dirty_flag()
                     self.update_info_box()
@@ -1186,9 +1211,9 @@ class TaggerWindow(QtWidgets.QMainWindow):
             self.adjust_load_style_combo()
             self.cbLoadDataStyle.currentIndexChanged.connect(self.set_load_data_style)
 
-    def set_save_data_style(self, s: str) -> None:
-        self.save_data_style = self.cbSaveDataStyle.itemData(s)
-        self.config[0].internal__save_data_style = self.save_data_style
+    def set_save_data_style(self) -> None:
+        self.save_data_styles = self.cbSaveDataStyle.currentData()
+        self.config[0].internal__save_data_style = self.save_data_styles
         self.update_metadata_style_tweaks()
         self.update_menus()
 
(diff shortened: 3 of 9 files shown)


comictaggerlib/pagelisteditor.py
```diff
@@ -369,11 +369,12 @@ def emit_front_cover_change(self) -> None:
...
             self.firstFrontCoverChanged.emit(self.first_front_page)
 
-    def set_metadata_style(self, data_style: str) -> None:
+    def set_metadata_style(self, data_styles: list[str]) -> None:
         # depending on the current data style, certain fields are disabled
-        if data_style:
-            self.metadata_style = data_style
+        if data_styles:
+            styles = [metadata_styles[style] for style in data_styles]
+            enabled_widgets = [attr for style in styles for attr in style.supported_attributes]
```
lordwelch: This needs to be put into a proper for loop. Yes it's faster but it takes me longer than 10 seconds to figure out what is happening on this line

lordwelch:
I'll test this out tomorrow.... probably

mizaki:
Also, with the possible (seems like mostly dark style), when the read style is the same `CheckableComboBox` the look will be more unified as it kind of works if the read/write style widget look the same and the source one is a little different.

mizaki:
Also, also is the question of converting over settings. As it's alpha should users be expected to remove/change the save style in the `settings.json`? Otherwise, it's leaving in code? Maybe change it over programatically and mark the code for removal at a later date?

comictaggerlib/taggerwindow.py
```diff
@@ -1211,14 +1237,15 @@ def update_metadata_credit_colors(self) -> None:
 
     def update_metadata_style_tweaks(self) -> None:
         # depending on the current data style, certain fields are disabled
-
-        enabled_widgets = metadata_styles[self.save_data_style].supported_attributes
+        enabled_widgets = [
```
lordwelch: Same as before write out the for loop here.

comictaggerlib/taggerwindow.py
```diff
@@ -1791,36 +1833,40 @@ def identify_and_tag_single_archive(
...
+                        self.auto_tag_log(f"{metadata_styles[style].name()} save failed ;-(\n")
+                    else:
+                        match_results.good_matches.append(
+                            Result(
+                                Action.save,
+                                Status.success,
+                                ca.path,
+                                online_results=matches,
+                                match_status=MatchStatus.good_match,
+                            )
```
lordwelch: This should break the loop. Operating on files that are potentially corrupt can cause further data loss

comictaggerlib/taggerwindow.py
```diff
@@ -1791,36 +1833,40 @@ def identify_and_tag_single_archive(
...
                         )
-                    )
-                    success = True
-                    self.auto_tag_log("Save complete!\n")
-                ca.load_cache([self.load_data_style, self.save_data_style])
+                        success = True
+                        self.auto_tag_log(f"{metadata_styles[style].name()} save complete!\n")
+
+                for style in self.save_data_styles:
```
lordwelch: Why is this in a loop? The function takes a list.

comictaggerlib/ui/taggerwindow.ui
```diff
@@ -59,30 +59,13 @@
...
-             </widget>
-            </item>
-            <item row="2" column="1">
-             <widget class="QComboBox" name="cbSaveDataStyle"/>
-            </item>
             <item row="0" column="0">
              <widget class="QLabel" name="lbl_md_source">
               <property name="text">
-               <string>Metadata Source</string>
+               <string>Data Source</string>
```
lordwelch, line 68: Why was this renamed?

comictaggerlib/taggerwindow.py
```diff
@@ -214,16 +214,31 @@ def __init__(
...
+        self.save_data_styles = config[0].internal__save_data_style
         self.load_data_style = config[0].internal__load_data_style
 
+        # Add multiselect combobox
+        self.cbSaveDataStyle = CheckableComboBox()
+        self.cbSaveDataStyle.setToolTip("At least 1 save style is required")
+        # Add normal combobox for read style (TODO support multiple read styles)
+        self.cbLoadDataStyle = QtWidgets.QComboBox()
+
+        # Need to set minimum or formLayout will resize larger than 230px (info and cover width)
```
lordwelch: What is "info and cover width"

comictaggerlib/taggerwindow.py
```diff
@@ -214,16 +214,31 @@ def __init__(
...
+        self.cbLoadDataStyle.setMinimumWidth(100)
+        self.cbSaveDataStyle.setMinimumWidth(100)
+        self.formLayout.addRow("Read Style", self.cbLoadDataStyle)
+        self.formLayout.addRow("Modify Styles", self.cbSaveDataStyle)
+        # Gets overwritten from ui file after using addRow
+        self.formLayout.setLabelAlignment(QtCore.Qt.AlignRight)
```
lordwelch: Give formLayout an actual name.

comictaggerlib/taggerwindow.py
```diff
@@ -1460,7 +1488,8 @@ def populate_combo_boxes(self) -> None:
         self.cbFormat.addItem("Year One")
 
     def remove_auto(self) -> None:
-        self.remove_tags(self.save_data_style)
+        for style in self.save_data_styles:
+            self.remove_tags(style)
```
lordwelch: remove_tags needs to take a list of styles instead of doing this. This will cause several popups to happen for every style selected

comictaggerlib/taggerwindow.py
```diff
@@ -214,16 +214,31 @@ def __init__(
...
+        self.cbSaveDataStyle.setToolTip("At least 1 save style is required")
+        # Add normal combobox for read style (TODO support multiple read styles)
+        self.cbLoadDataStyle = QtWidgets.QComboBox()
+
+        # Need to set minimum or formLayout will resize larger than 230px (info and cover width)
+        self.cbLoadDataStyle.setMinimumWidth(100)
+        self.cbSaveDataStyle.setMinimumWidth(100)
+        self.formLayout.addRow("Read Style", self.cbLoadDataStyle)
+        self.formLayout.addRow("Modify Styles", self.cbSaveDataStyle)
+        # Gets overwritten from ui file after using addRow
```
lordwelch: How does this get overwritten from the ui file? It is loaded before this on line 89

lordwelch (changes requested):
The checkable combobox is a decent idea. But currently clicking on the text does not consistently bring up the list on macos

comictaggerlib/ui/taggerwindow.ui
mizaki: I renamed it because it takes up too much space and means the more important names of the source or styles get more truncated.

comictaggerlib/taggerwindow.py
mizaki: That is the information box and cover image under the source and styles form. I'll add a bit more detail to the message.

comictaggerlib/taggerwindow.py
mizaki: Good question. I think I was thinking it was a list comprehension...

comictaggerlib/taggerwindow.py
mizaki: It turns out even though I had saved with AlignRight (as it was before), because it was the default/unchanged it didn't save it and then defaulted to AlignLeft (which appears to still be incorrect as KDE should be AlignRight by default in the docs as I read it...). I cycled it between AlignLeft and AlignRight and then it saved.

mizaki:
> The checkable combobox is a decent idea. But currently clicking on the text does not consistently bring up the list on macos

For me on Linux with KDE, I can spam click the text and it works as expected. Does it just highlight the text?
spam_click_write_style

lordwelch:
They did some questionable things, which I've now fixed.
This should work better across multiple platforms, and at least on macOS it would previously change the style which it now does not
```python
    # Multiselect combobox from: https://gis.stackexchange.com/a/351152 (with custom changes)
    class CheckableComboBox(QtWidgets.QComboBox):
        itemChecked = pyqtSignal(str, bool)

        def __init__(self, *args: Any, **kwargs: Any):
            super().__init__(*args, **kwargs)
            # Prevent popup from closing when clicking on an item
            self.view().viewport().installEventFilter(self)

            # Keeps track of when the combobox list is shown
            self.justShown = False

        def resizeEvent(self, event: Any) -> None:
            # Recompute text to elide as needed
            super().resizeEvent(event)
            self._updateText()

        def eventFilter(self, obj: Any, event: Any) -> bool:
            # Allow events befor the combobox list is shown
            if obj == self.view().viewport():
                # We record that the combobox list has been shown
                if event.type() == QEvent.Show:
                    self.justShown = True
                # We record that the combobox list has hidden,
                # this will happen if the user does not make a selection
                # but clicks outside of the combobox list or presses escape
                if event.type() == QEvent.Hide:
                    self._updateText()
                    self.justShown = False
                # QEvent.MouseButtonPress is inconsistent on activation because double clicks are a thing
                if event.type() == QEvent.MouseButtonRelease:
                    # If self.justShown is true it means that they clicked on the combobox to change the checked items
                    # This is standard behavior (at least on macos) but I think it is surprising when it has a multiple select
                    if self.justShown:
                        self.justShown = False
                        return True

                    # Find the current index and item
                    index = self.view().indexAt(event.pos())
                    self.toggleItem(index.row())
                    return True
            return False

        def currentData(self) -> list[Any]:
            # Return the list of all checked items data
            res = []
            for i in range(self.count()):
                item = self.model().item(i)
                if item.checkState() == Qt.Checked:
                    res.append(self.itemData(i))
            return res

        def addItem(self, text, data=None):
            super().addItem(text, data)
            if self.count() == 1:
                self.model().item(0).setCheckState(Qt.Checked)

        def onDataChanged(self, index: Any) -> None:
            item = self.model().item(index.row())
            # If we have at least one item checked emit itemChecked with the current check state and update text
            # Require at least one item to be checked and provide a tooltip

        def _updateText(self) -> None:
            texts = []
            for i in range(self.count()):
                item = self.model().item(i)
                if item.checkState() == Qt.Checked:
                    texts.append(item.text())
            text = ", ".join(texts)

            # Compute elided text (with "...")

            # The QStyleOptionComboBox is needed for the call to subControlRect
            so = QtWidgets.QStyleOptionComboBox()
            # init with the current widget
            so.initFrom(self)

            # Ask the style for the size of the text field
            rect = self.style().subControlRect(QtWidgets.QStyle.CC_ComboBox, so, QtWidgets.QStyle.SC_ComboBoxEditField)

            # Compute the elided text
            elidedText = self.fontMetrics().elidedText(text, Qt.ElideRight, rect.width())

            # This CheckableComboBox does not use the index, so we clear it and set the placeholder text
            self.setCurrentIndex(-1)
            self.setPlaceholderText(elidedText)

        def setItemChecked(self, index: Any, state: bool) -> None:
            qt_state = Qt.Checked if state else Qt.Unchecked
            item = self.model().item(index)
            current = self.currentData()
            # Require at least one selection
            if len(current) == 1 and not state and item.checkState() == Qt.Checked:
                QtWidgets.QToolTip.showText(QtGui.QCursor.pos(), self.toolTip(), self, QRect(), 3000)
                return

            if len(current) > 0:
                item.setCheckState(qt_state)
                self.itemChecked.emit(self.itemData(index), state)
                self._updateText()

        def toggleItem(self, index: int) -> None:
            if self.model().item(index).checkState() == Qt.Checked:
                self.setItemChecked(index, Qt.Unchecked)
            else:
                self.setItemChecked(index, Qt.Checked)
```

https://github.com/comictagger/comictagger/assets/7547075/616d7b95-fefb-493d-907a-1d8fc49a17ec


https://github.com/comictagger/comictagger/assets/7547075/e272121b-3e4e-450e-8e4f-b63086758d54

mizaki:
I believe that should be everything covered now?

I don't know why the hash has changed for the Cory cover image with the test.

comictaggerlib/pagelisteditor.py
```diff
@@ -370,11 +370,16 @@ def emit_front_cover_change(self) -> None:
...
+    def set_metadata_style(self, data_styles: list[str]) -> None:
         # depending on the current data style, certain fields are disabled
-        if data_style:
-            self.metadata_style = data_style
+        if data_styles:
+            styles = [metadata_styles[style] for style in data_styles]
+            enabled_widgets = []
+            for style in styles:
+                for attr in style.supported_attributes:
+                    enabled_widgets.append(attr)
```
lordwelch: Lets change enabled_widgets to a set and you can can remove the inner loop:
```suggestion
        if data_styles:
            styles = [metadata_styles[style] for style in data_styles]
            enabled_widgets = set()
            for style in styles:
                enabled_widgets.update(style.supported_attributes)
```

comictaggerlib/taggerwindow.py
```diff
@@ -1155,12 +1170,22 @@ def commit_metadata(self) -> None:
                 QtWidgets.QApplication.setOverrideCursor(QtGui.QCursor(QtCore.Qt.CursorShape.WaitCursor))
                 self.form_to_metadata()
 
-                success = self.comic_archive.write_metadata(self.metadata, self.save_data_style)
+                fail_list = []
+                # Save each style
+                for style in self.save_data_styles:
+                    success = self.comic_archive.write_metadata(self.metadata, style)
+                    if not success:
+                        fail_list.append(style)
```
lordwelch: This should break, I'm not sure what happened you had fixed this earlier

comictaggerlib/autotagmatchwindow.py
```diff
@@ -247,10 +247,10 @@ def save_match(self) -> None:
...
         md.overlay(ct_md)
-        success = ca.write_metadata(md, self._style)
-        ca.load_cache(list(metadata_styles))
-
-        QtWidgets.QApplication.restoreOverrideCursor()
+        for style in self._styles:
+            success = ca.write_metadata(md, style)
+            QtWidgets.QApplication.restoreOverrideCursor()
+            if not success:
+                QtWidgets.QMessageBox.warning(self, "Write Error", "Saving the tags to the archive seemed to fail!")
```
lordwelch: This needs to break on failure

comictaggerlib/taggerwindow.py
```diff
@@ -1186,23 +1211,28 @@ def set_load_data_style(self, s: str) -> None:
...
 
     def update_metadata_credit_colors(self) -> None:
-        style = metadata_styles[self.save_data_style]
-        enabled = style.supported_attributes
+        styles = [metadata_styles[style] for style in self.save_data_styles]
+        enabled = []
+
+        for style in styles:
+            for attr in style.supported_attributes:
+                enabled.append(attr)
```
lordwelch: Same as above, turn this into a set and remove the inner loop

comictaggerlib/taggerwindow.py
```diff
@@ -1212,14 +1242,17 @@ def update_metadata_credit_colors(self) -> None:
 
     def update_metadata_style_tweaks(self) -> None:
         # depending on the current data style, certain fields are disabled
+        enabled_widgets = []
+        for style in self.save_data_styles:
+            for attr in metadata_styles[style].supported_attributes:
+                enabled_widgets.append(attr)
```
lordwelch: Same as above, turn this into a set and remove the inner loop

comictaggerlib/taggerwindow.py
```diff
@@ -1508,20 +1552,23 @@ def remove_tags(self, style: str) -> None:
...
                     QtCore.QCoreApplication.processEvents()
-                    if ca.has_metadata(style) and ca.is_writable():
-                        if not ca.remove_metadata(style):
-                            failed_list.append(ca.path)
-                        else:
-                            success_count += 1
-                        ca.load_cache(list(metadata_styles))
+                    for style in styles:
+                        if ca.has_metadata(style) and ca.is_writable():
+                            if not ca.remove_metadata(style):
```
lordwelch: Lets invert the check here, negatives almost always make things more difficult to understand.

comictaggerlib/taggerwindow.py
```diff
@@ -1508,20 +1552,23 @@ def remove_tags(self, style: str) -> None:
...
+                                failed_list.append(ca.path)
+                                # Abandon any further tag removals to prevent any greater damage to archive
+                                break
+                            else:
+                                success_count += 1
+                            ca.load_cache(list(metadata_styles))
```
lordwelch: lets also reset the cache before loading the cache

comictaggerlib/taggerwindow.py
```diff
@@ -1538,9 +1585,13 @@ def copy_tags(self) -> None:
         has_src_count = 0
 
         src_style = self.load_data_style
-        dest_style = self.save_data_style
+        dest_styles = list(self.save_data_styles)
```
lordwelch: Is there a reason for the list call here?

comictaggerlib/taggerwindow.py
```diff
@@ -1580,26 +1634,32 @@ def copy_tags(self) -> None:
...
+                            if prog_dialog.wasCanceled():
+                                break
+
+                            prog_dialog.setValue(prog_idx)
+                            prog_dialog.setLabelText(str(ca.path))
+                            center_window_on_parent(prog_dialog)
+                            QtCore.QCoreApplication.processEvents()
+
+                        if ca.has_metadata(src_style) and ca.is_writable():
+                            md = ca.read_metadata(src_style)
```
lordwelch: This needs to be moved out of the loop, no reason to read the same metadata several times

comictaggerlib/taggerwindow.py
```diff
@@ -1580,26 +1634,32 @@ def copy_tags(self) -> None:
...
-                        if not ca.write_metadata(md, dest_style):
-                            failed_list.append(ca.path)
-                        else:
-                            success_count += 1
+                            if not ca.write_metadata(md, style):
+                                failed_list.append(ca.path)
+                            else:
+                                if not ca_saved:
+                                    success_count += 1
+                                    ca_saved = True
```
lordwelch: You have an extra indent here

comictaggerlib/taggerwindow.py
```diff
@@ -1580,26 +1634,32 @@ def copy_tags(self) -> None:
...
 
-                        ca.load_cache([self.load_data_style, self.save_data_style, src_style, dest_style])
+                        ca.load_cache([self.load_data_style, self.save_data_style, src_style, style])
```
lordwelch: this needs a de-dent, reset the cache before loading and the cache only needs to include `self.load_data_style` and `*self.save_data_styles`

comictaggerlib/taggerwindow.py
```diff
@@ -1792,36 +1852,42 @@ def identify_and_tag_single_archive(
...
+                for style in self.save_data_styles:
+                    if not ca.write_metadata(md, style):
+                        match_results.write_failures.append(
+                            Result(
+                                Action.save,
+                                Status.write_failure,
+                                ca.path,
+                                online_results=matches,
+                                match_status=MatchStatus.good_match,
+                            )
```
lordwelch: There should be one result per comic archive and use these attributes, just like how the cli does
```python
    tags_deleted: list[str] = dataclasses.field(default_factory=list)
    tags_written: list[str] = dataclasses.field(default_factory=list)
```

comictaggerlib/taggerwindow.py
```diff
@@ -1792,36 +1852,42 @@ def identify_and_tag_single_archive(
...
+                                Status.success,
+                                ca.path,
+                                online_results=matches,
+                                match_status=MatchStatus.good_match,
+                            )
+                        )
+                        success = True
+                        self.auto_tag_log(f"{metadata_styles[style].name()} save complete!\n")
+
+                ca.load_cache([self.load_data_style] + self.save_data_styles)
```
lordwelch, line 1882: Lets reset the cache before loading the cache

comictaggerlib/taggerwindow.py
```diff
@@ -1792,36 +1852,42 @@ def identify_and_tag_single_archive(
...
-                    match_results.write_failures.append(
-                        Result(
-                            Action.save,
-                            Status.write_failure,
-                            ca.path,
-                            online_results=matches,
-                            match_status=MatchStatus.good_match,
+                # Save each style
+                for style in self.save_data_styles:
+                    if not ca.write_metadata(md, style):
```
lordwelch: Lets invert the match here as well

comictaggerlib/taggerwindow.py
```diff
@@ -215,15 +216,29 @@ def __init__(
...
+        self.cbSaveDataStyle.setToolTip("At least 1 save style is required")
+        # Add normal combobox for read style (TODO support multiple read styles)
+        self.cbLoadDataStyle = QtWidgets.QComboBox()
+
+        # Need to set minimum or source_style_formLayout will resize larger than 230px which will affect the
+        # file info box and cover image width underneath
+        self.cbLoadDataStyle.setMinimumWidth(100)
+        self.cbSaveDataStyle.setMinimumWidth(100)
+        self.source_style_formLayout.addRow("Read Style", self.cbLoadDataStyle)
+        self.source_style_formLayout.addRow("Modify Styles", self.cbSaveDataStyle)
```
lordwelch: We can actually put all of this into the ui file.
Right click on the widget and click on promote
Screenshot 2024-02-10 at 1 22 52 PM

comictaggerlib/ui/taggerwindow.ui
```diff
@@ -52,37 +52,23 @@
         <item>
          <layout class="QVBoxLayout" name="verticalLayout">
           <item>
-           <layout class="QFormLayout" name="formLayout">
+           <layout class="QFormLayout" name="source_style_formLayout">
```
lordwelch: The comment on taggerwindow lines +230 to +241 removes the need for this to be named but.
`style_layout` is a much more manageable name

lordwelch (changes requested):
The hash change was my last commit, I'll fix the test.

comictaggerlib/taggerwindow.py
mizaki: Me too. I must have lost it in the squish...

comictaggerlib/autotagmatchwindow.py
mizaki: I'll add the style name too otherwise it's going to be too vague.

comictaggerlib/taggerwindow.py
mizaki: No...

comictaggerlib/taggerwindow.py
mizaki: Should the `reset_cache` and `load_cache` be outside the loop?

comictaggerlib/taggerwindow.py
mizaki: I believe this is correct now? I didn't see where any `tags_deleted` would come from?

comictaggerlib/taggerwindow.py
mizaki: I thought it would require a C header. Should have looked into it further...

comictaggerlib/taggerwindow.py
lordwelch: Outside of the style loop

comictaggerlib/taggerwindow.py
lordwelch: You also need to call continue when this check fails otherwise md may not exist when it tries to copy the tags

lordwelch (changes requested):
The last thing I want to see is an option to use the ID instead of the full name for the checkable combo box so that I can actually see more than one at a time. It should be a file only option

mizaki:
> The last thing I want to see is an option to use the ID instead of the full name for the checkable combo box so that I can actually see more than one at a time. It should be a file only option

Under the `General` tab? With the multi-read style I think a "Metadata Style" tab might be needed so I could add that now with that option?

mizaki:
> > The last thing I want to see is an option to use the ID instead of the full name for the checkable combo box so that I can actually see more than one at a time. It should be a file only option
> 
> Under the `General` tab? With the multi-read style I think a "Metadata Style" tab might be needed so I could add that now with that option?

`Metadata Types` tab could be split up with a separator and the styles placed in there? I think a `listwidget` with the styles and checkboxes, so same as the combobox but in a list instead of a dropdown.

mizaki:
> The last thing I want to see is an option to use the ID instead of the full name for the checkable combo box so that I can actually see more than one at a time. It should be a file only option

I set it to `False` by default but did consider `True` however showing the "friendly" name first made more sense.

When the read multi-styles is done I'll add the option to change the metadata styles in the settings window too.

The new function `populate_style_names` can be used for the read style when it goes in.

I'm not sure if there is a better way to go about updating the style names when the option has been checked?

lordwelch (approved):

Issue #587, opened 2024-01-08
Removing credits from CLI


atimeforwolves:
When using ComicTagger with Comic Vine as metadata source, I am getting duplicates in credits with one having a number in []. This ends up bogging down my view through Kavita as when viewing an entire volume it populates all unique entries in the credits. As such, I'd rather just remove the credit entries entirely. Is there a way to do this from the cli? I'm not having any luck with anything I try after -m

Ex:
Writer: John Broome
Writer: John Broome [48152]

edit: regarding Comic Info XML (cix)

lordwelch:
`-m credits=` should work

atimeforwolves:
> `-m credits=` should work

Thank you! I had tried this earlier but maybe I had used credit instead of credits and didn't realize.

atimeforwolves:
It looks like even though it did not error out, the credits are still there after checking: 
![image](https://github.com/comictagger/comictagger/assets/105068114/6e1eb176-badd-4f87-929a-a4a9ac341427)

For clarity, I have been using this batch script, which has been working for all other metatags:
for /r %%A IN (\*.cb*) do ( "full\path\to\comictagger.exe" -s -t cix -m "credits=" "%%~A") pause

atimeforwolves:
It looks like this is how it's being assigned, but I'm not sure on the syntax:
```
    def overlay_credits(self, new_credits: list[Credit]) -> None:
        for c in new_credits:
            primary = bool("primary" in c and c["primary"])

            # Remove credit role if person is blank
            if c["person"] == "":
                for r in reversed(self.credits):
                    if r["role"].casefold() == c["role"].casefold():
                        self.credits.remove(r)
            # otherwise, add it!
            else:
                self.add_credit(c["person"], c["role"], primary)
```
Doesn't seem like it's able to handle removing credit metadata?

PR #598, opened 2024-02-04
Add overlay modes for metadata; overwrite, add missing etc.

mizaki:
This is the barebones of some overlay options. It's missing the settings which I will add if you are happy with the general idea. It's also missing any tests for the new options. I have done nothing with the `notes` as yet.

When it comes to the notes, using `CVDB xxxxx` will still work with Mylar3 and as long as `Issue ID` is not used (such as `Metron Issue ID` and instead `Metron ID`) then the multi-source `Tagged with CT etc. etc. CVDB xxxxx, GCD xxxxxx, Metron xxxxx` will not cause problems. I've checked Komga, Kavita and LANraragi and none appear to use the `notes` field.


comicapi/genericmetadata.py
```diff
@@ -103,6 +103,13 @@ class GenericMetadata:
     editor_synonyms = ("editor", "edits", "editing")
     _translator_synonyms = ("translator", "translation")
 
+    overlay_original = 0
+    overlay_overwrite = 1
+    overlay_overwrite_no_clear = 2
+    overlay_add_missing = 3
+    overlay_combine_new = 4
+    overlay_combine_cur = 5
```
lordwelch: Make an enum class for this. At a minimum.

comicapi/genericmetadata.py
```diff
@@ -202,25 +209,112 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
         new_md.__post_init__()
         return new_md
 
-    def overlay(self, new_md: GenericMetadata) -> None:
-        """Overlay a metadata object on this one
-
-        That is, when the new object has non-None values, over-write them
-        to this one.
+    def casefold_check(self, a: str, b: str) -> bool:
```
lordwelch: This doesn't help anything. remove it and use the comparison directly.

comicapi/genericmetadata.py
```diff
@@ -202,25 +209,112 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        return a.casefold() == b.casefold()
+
+    def overlay(self, new_md: GenericMetadata, mode: int = 0) -> None:
```
lordwelch: Never use "magic numbers" in any functions always use a named variable. It shouldn't matter if this is a int, float, str or a class this should be a name to refer to the mode that happens.

comicapi/genericmetadata.py
```diff
@@ -202,25 +209,112 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        3. Add Missing - Any values that are None or an empty str/list/set, add them from 'new' to 'self'.
+        4. Combine New - Where possible combine the values of 'self' with the values of 'new'. Any values that collide
+        the 'new' value is used (ignoring non-values such as empty strings, int as 0, etc.).
+        5. Combine Current - Same as 2 but 'self' value is kept with collisions.
         """
 
         def assign(cur: str, new: Any) -> None:
-            if new is not None:
-                if isinstance(new, str) and len(new) == 0:
+            if mode == self.overlay_original:
```
lordwelch: Don't do this. You are going 8-levels deep on some of these. At a minimum make each of these separate functions that get called here.

You could also just make the mode argument the actual function so you don't even need the if chain

comicapi/genericmetadata.py
```diff
@@ -278,20 +372,65 @@ def assign(cur: str, new: Any) -> None:
         assign("_cover_image", new_md._cover_image)
         assign("_alternate_images", new_md._alternate_images)
 
-    def overlay_credits(self, new_credits: list[Credit]) -> None:
-        if isinstance(new_credits, str) and len(new_credits) == 0:
+    def overlay_credits(self, new_credits: list[Credit], mode: int = 0) -> None:
```
lordwelch: Same as above with assign

comicapi/genericmetadata.py
mizaki: There is the Enum https://docs.python.org/3/library/enum.html which I looked at but thought I'd replicate what had been done in `IssueIdentifier` etc. Should I use the enum class or just make a normal class?

lordwelch: Use the enum class. ComicTagger was started back on python 2.7 and still has some leftover concepts from back then

mizaki:
I've tried to simplify things with https://github.com/comictagger/comictagger/pull/598/commits/ddbc5d83f8d65c0dbb5687b92ca400ef4202471f (just the `assign_` changes for now) I didn't know if it's better check if a string is 0 rather that just set the `cur` to whatever `new` is even if it was an empty string?

mizaki:
When it comes to saving the setting, is it best to use the Enum number or name?

lordwelch:
> When it comes to saving the setting, is it best to use the Enum number or name?

parser.add_setting("value", default=ENUM.VALUE, type=ENUM, cmdline=False)

This won't work with int values with our current file loading code.
We would need ` or isinstance(default, Enum) or (isinstance(setting.type, type) and issubclass(setting.type, Enum))` added to https://github.com/comictagger/comictagger/blob/09490b8ebf70415211dc856663ea1c1fd258b47b/comictaggerlib/ctsettings/__init__.py#L58

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+            remove_set = set()
+            for a in getattr(self, cur):
+                for b in new:
+                    if a.casefold() == b.casefold():
+                        remove_set.add(b)
+                        break
+            new.difference_update(remove_set)
+            getattr(self, cur).update(new)
+
+    def assign_original(self, cur: str, new: Any) -> None:
```
lordwelch: This should be `assign_overlay`

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+                if isinstance(getattr(self, cur), (list, set)):
+                    getattr(self, cur).clear()
+                else:
+                    setattr(self, cur, None)
+            elif isinstance(new, (list, set)) and len(new) == 0:
+                pass
+            else:
+                setattr(self, cur, new)
+
+    def assign_overwrite(self, cur: str, new: Any) -> None:
```
lordwelch: If everything is being overwritten isn't it just making a copy of the new metadata object?

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        if isinstance(new, (str, list, set)) and len(new) == 0:
+            if isinstance(getattr(self, cur), (list, set)):
+                getattr(self, cur).clear()
+            else:
+                setattr(self, cur, None)
+        else:
+            setattr(self, cur, new)
+
+    def assign_overwrite_no_clear(self, cur: str, new: Any) -> None:
```
lordwelch: What's the point in this? This also allows a string to be assigned to a list

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+                setattr(self, cur, new)
+            elif isinstance(new, (bool, int, float)):
+                setattr(self, cur, new)
+
+    def assign_add_missing(self, cur: str, new: Any) -> None:
+        if new is not None and not getattr(self, cur):
+            setattr(self, cur, new)
+
+    def assign_combine_new(self, cur: str, new: Any) -> None:
+        # Combine lists and sets and prefer 'new' for strings, floats and ints
```
lordwelch: So overlay + merge list/set?

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        if new is not None:
+            if isinstance(new, (list, set)) and len(new) > 0 and isinstance(getattr(self, cur), (list, set)):
+                # Check for dupes with casefold to replace accented chars
+                self.assign_dedupe_new(new, cur)
+            elif isinstance(new, str) and len(new) > 0:
+                setattr(self, cur, new)
+            elif isinstance(new, (int, float)) and new > 0:
+                setattr(self, cur, new)
```
lordwelch: Why is 0 ignored? `None` is used for empty. This makes it impossible to update an issue number 0 such as https://www.dc.com/comics/batman-2011/batman-0

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+
+    def assign_combine_cur(self, cur: str, new: Any) -> None:
+        # Combine lists and sets and prefer 'self' for strings, floats and ints
+        if new is not None:
+            if isinstance(new, str) and (len(getattr(self, cur)) == 0 or getattr(self, cur) is None):
+                setattr(self, cur, new)
+            elif isinstance(new, (list, set)) and len(new) > 0 and isinstance(getattr(self, cur), (list, set)):
+                self.assign_dedupe_cur(new, cur)
+            elif isinstance(new, (int, float)) and getattr(self, cur) == 0:
```
lordwelch: This won't update anything 90% of the time int attributes do not start-off as 0 they start as `None`

comicapi/genericmetadata.py
```diff
@@ -202,24 +212,122 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+                if getattr(self, cur) is None:
+                    setattr(self, cur, new)
 
-        That is, when the new object has non-None values, over-write them
-        to this one.
+    def overlay(self, new_md: GenericMetadata, mode: Enum = OverlayMode.ORIGINAL) -> None:
+        """Overlay a new metadata object on this one
+
+        Modes:
+        0. Original behaviour: That is, when the new object has non-None values, over-write them to this one.
```
lordwelch: Don't use numbers here

mizaki:
Is https://github.com/comictagger/comictagger/pull/598/commits/75c825be1d880e740f3b562a66b7b6b6c07e2cd5 in the right area with the class?

lordwelch (commented):
Add a single test for each mode that covers each type so that we know what it actually does, it will also make some of the mistakes a little more obvious.

Comments are in the wrong place now but still apply.

comicapi/genericmetadata.py
mizaki: I don't know if it was supposed to do something else originally but yes, it will now overwrite everything. I think this should then replace the `Clear all existing metadata before import...` so everything is in one location/setting?

comicapi/genericmetadata.py
mizaki: That makes issue number a special case? Is there anything else that could be considered that you'd want to overwrite a value `> 0` with `0`? 

`None` is for empty but a talker might return a 0 for `null/None`, is the talker expected to decide if `0` should be `None`?

comicapi/genericmetadata.py
lordwelch: We can look at that after this pr

comicapi/genericmetadata.py
lordwelch: Yes, if the value is different from the default (None for most types) it's assumed that it was set on purpose

lordwelch: The current condition also disallows setting any negative numbers. The whole reason int values default to None is so that all integer values can be used

mizaki:
I'm trying to understand the current overlay:
``` python
    def assign_overlay(self, cur: str, new: Any) -> None:
        """Overlay - When the new object has non-None values, over-write them to this one."""
        if new is not None:
            if isinstance(new, str) and len(new) == 0:
                if isinstance(getattr(self, cur), (list, set)): # If an empty string is passed to a list/set field then it is cleared
                    getattr(self, cur).clear()
                else:
                    setattr(self, cur, None) # Otherwise it's any empty string in a string field and should set to None
            elif isinstance(new, (list, set)) and len(new) == 0: # If an empty list/set is passed, do nothing. Why does this not clear the field the same as an empty string?
                pass
            else:
                setattr(self, cur, new) # No type checking, just set the `cur` to `new`
```
Should an empty list/set not `clear` it? Am I missing something?

And in general, should `assign` be doing type checking?

lordwelch:
An empty list is the default

comicapi/genericmetadata.py
mizaki: Yes (mostly?). It will merge lists/sets and replace all other values with `new` but won't clear a list/set on an empty string. Is "merge" the preferred nomenclature?

mizaki:
> An empty list is the default

But should a list/set expect an empty string? That should be a type error?

mizaki:
I've added some tests and given each one a doc string so hopefully it makes more sense now. Credit tests are left to do.

lordwelch:
> But should a list/set expect an empty string? That should be a type error?

I believe this was set in order to make the `-m` option work for clearing lists

comicapi/genericmetadata.py
lordwelch: Here's an example of me doing it with strings and breaking something
https://github.com/comictagger/comictagger/blob/09490b8ebf70415211dc856663ea1c1fd258b47b/comictaggerlib/ctsettings/plugin.py#L65
It should be `Archive` is it is set here
https://github.com/comictagger/comictagger/blob/09490b8ebf70415211dc856663ea1c1fd258b47b/comictaggerlib/ctsettings/plugin.py#L100

mizaki:
I had to duplicate the `StrEnum` code from `resulttype.py` as using it caused an import loop. I guess the best thing to do is move it to a new/another file, if so, which one?

lordwelch:
> I had to duplicate the `StrEnum` code from `resulttype.py` as using it caused an import loop. I guess the best thing to do is move it to a new/another file, if so, which one?

Copying it is fine, it'll get removed eventually when CT no longer supports 3.10

comicapi/genericmetadata.py
```diff
@@ -202,24 +252,129 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+                setattr(self, cur, new)
+
+    def overlay(self, new_md: GenericMetadata, mode: OverlayMode = OverlayMode.overlay) -> None:
+        """Overlay a new metadata object on this one"""
+
+        assign_funcs = {}
+        for mem in OverlayMode:
+            assign_funcs[mem.name] = getattr(self, "assign_" + mem.name)
 
         def assign(cur: str, new: Any) -> None:
```
lordwelch: There's a whole lot of processing going on here that doesn't need to happen. This dict can just be manually created.

Instead of doing any of that how about this
```python
def assign_overlay(md: GenericMetadata, attribute: str, new_value: Any) -> None:
    """Overlay - When the new object has non-None values, over-write them to this one."""
    if new_value is not None:
        if isinstance(new_value, str) and len(new_value) == 0:
            if isinstance(getattr(md, attribute), (list, set)):
                getattr(md, attribute).clear()
            else:
                setattr(md, attribute, None)
        elif isinstance(new_value, (list, set)) and len(new_value) == 0:
            pass
        else:
            setattr(md, attribute, new_value)
...
class GenericMetadata:
    ...
    def overlay(self, new_md: GenericMetadata, assign=assign_overlay) -> None:
        """Overlay a new metadata object on this one"""

        if not new_md.is_empty:
            self.is_empty = False

        assign(self, "tag_origin", new_md.tag_origin)
```
Or instead of keeping the current function signature
```python
    def overlay(self, new_md: GenericMetadata, assign=assign_overlay) -> None:
        """Overlay a new metadata object on this one"""

        if not new_md.is_empty:
            self.is_empty = False

        self.tag_origin = assign(self.tag_origin, new_md.tag_origin)
```

tests/genericmetadata_test.py
```diff
@@ -22,6 +23,78 @@ def test_metadata_overlay(md: comicapi.genericmetadata.GenericMetadata, replaced
...
 
 
+def test_metadata_overlay_overwrite():
+    md = comicapi.genericmetadata.GenericMetadata(series="test", issue="1", title="test", genres={"test", "test2"})
+    overwrite = comicapi.genericmetadata.GenericMetadata(
+        series="test2", issue="2", title="test2", genres={"test3", "test4"}
+    )
+    md.overlay(overwrite, OverlayMode.overwrite)
+
+    assert md == overwrite
```
lordwelch: This test makes it obvious that this mode is useless. Just use md.copy()

tests/genericmetadata_test.py
```diff
@@ -22,6 +23,78 @@ def test_metadata_overlay(md: comicapi.genericmetadata.GenericMetadata, replaced
...
+    )
+    overwrite = comicapi.genericmetadata.GenericMetadata(
+        series="test2", issue="2", title="test2", genres=set(), story_arcs=[]
+    )
+    expected = comicapi.genericmetadata.GenericMetadata(
+        series="test2", issue="2", title="test2", genres={"test", "test2"}, story_arcs=["arc"]
+    )
+    md.overlay(overwrite, OverlayMode.overwrite_no_clear)
+
+    assert md == expected
```
lordwelch: This test needs to show how it is different from overlay. I don't think it ever makes sense to use `overwrite_no_clear` as it is written

comicapi/genericmetadata.py
```diff
@@ -202,24 +252,129 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        if new is not None:
+            if isinstance(new, (list, set)) and len(new) > 0 and isinstance(getattr(self, cur), (list, set)):
+                setattr(self, cur, new)
+            elif isinstance(new, str) and len(new) > 0:
+                setattr(self, cur, new)
+            elif isinstance(new, (int, float, bool)):
+                setattr(self, cur, new)
+
+    def assign_add_missing(self, cur: str, new: Any) -> None:
+        """Add Missing - Any values that are None or an empty str/list/set, add them from 'new' to 'self'."""
```
lordwelch:
```suggestion
        """Add Missing - Any values that are NOT None or an empty str/list/set, add them from 'new' to 'self'."""
```

comicapi/genericmetadata.py
```diff
@@ -202,24 +252,129 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+
+    def assign_combine_cur(self, cur: str, new: Any) -> None:
+        """Combine (Current) - Same as Combine (New) but 'self' value is kept with collisions."""
+        if new is not None:
+            if isinstance(new, (list, set)) and isinstance(getattr(self, cur), (list, set)):
+                self.assign_dedupe(new, cur)
+            elif isinstance(new, str) and len(getattr(self, cur)) == 0:
+                setattr(self, cur, new)
+            elif getattr(self, cur) is None:
+                setattr(self, cur, new)
```
lordwelch: Why are there 2 of these. The only list/set type you account for is a str (because we don't have any others currently) but that means they are all strings! The only thing you are preserving is case. As named this should have the same implementation as overlay but instead of replacing a list it merges them `if new not in getattr(self, 'cur'): getattr(self, 'cur').add(new)` and `if new not in getattr(self, 'cur'): getattr(self, 'cur').append(new)`

tests/genericmetadata_test.py
```diff
@@ -22,6 +23,78 @@ def test_metadata_overlay(md: comicapi.genericmetadata.GenericMetadata, replaced
...
+    )
+    combine_cur = comicapi.genericmetadata.GenericMetadata(
+        series="test2", title="test2", genres={"test2", "test3", "test4"}, story_arcs=["arc1", "arc2"]
+    )
+    expected = comicapi.genericmetadata.GenericMetadata(
+        series="test", issue="1", title="test", genres={"test", "test2", "test3", "test4"}, story_arcs=["arc1", "arc2"]
+    )
+    md.overlay(combine_cur, OverlayMode.combine_cur)
+
+    assert md == expected
```
lordwelch: No, this is useless they can just reverse the call of `md.overlay(new_md)` to `new_md.overlay(md)`

lordwelch (changes requested):

comicapi/genericmetadata.py
lordwelch: This (`self.tag_origin = assign(self.tag_origin, new_md.tag_origin)`) now makes assign not representative of what it is doing so should be renamed if used

comicapi/genericmetadata.py
lordwelch: Just wanted to make sure I new what it was trying to do

tests/genericmetadata_test.py
lordwelch: As in anywhere that would have called `md.overlay(new_md, OverlayMode.overwrite)` should instead use `md = new_md.copy()`

lordwelch:
Also as a nitpick (because I don't remember to do this all the time anyways) you can change:
```python
        if new is not None:
            ...
```
to
```python
    if new is None:
        return None
    ...
```
to remove another indent level.

As it is now it's 10x better than the 8 levels of indent originally.

comicapi/genericmetadata.py
mizaki:
``` python
    def overlay(self, new_md: GenericMetadata, assign=assign_overlay) -> None:
        """Overlay a new metadata object on this one"""

        if not new_md.is_empty:
            self.is_empty = False

        self.tag_origin = assign(self.tag_origin, new_md.tag_origin)
```
The `overlay` gets called from outside and will pass in the `config[0].Comic_Book_Lover__metadata_overlay` for the `assign` so keeping `def overlay(self, new_md: GenericMetadata, mode: OverlayMode = OverlayMode.overlay) -> None:` makes the most sense? Along with:
```python
assign_funcs = {
            "overlay": self.assign_overlay,
            "overwrite": self.assign_overwrite,
            "overwrite_no_clear": self.assign_overwrite_no_clear,
            "add_missing": self.assign_add_missing,
            "combine_new": self.assign_combine_new,
            "combine_current": self.assign_combine_current,
        }

        def assign(cur: str, new: Any) -> None:
            assign_func = assign_funcs.get(mode.name, self.assign_overlay)
            assign_func(cur, new)
```

`self.tag_origin = assign(self.tag_origin, new_md.tag_origin)`
I did think having a return would be easier now there is multiple options, so I will change it to that.

mizaki:
Part of the problem with `overwrite` and `overwrite_no_clear` was me not understanding the issue of a string clearing a list/set.

Currently https://github.com/comictagger/comictagger/blob/09490b8ebf70415211dc856663ea1c1fd258b47b/comictaggerlib/ctsettings/types.py#L86 doesn't do any type conversions which it probably should here? 
https://github.com/comictagger/comictagger/blob/09490b8ebf70415211dc856663ea1c1fd258b47b/comictaggerlib/ctsettings/types.py#L132

Or, the `GenericMetadata.__post_init__` could do checks and conversions?
That way, overlay doesn't need to worry about the wrong type when replacing/adding/merging.

comicapi/genericmetadata.py
mizaki: > This (`self.tag_origin = assign(self.tag_origin, new_md.tag_origin)`) now makes assign not representative of what it is doing so should be renamed if used

Should `tag_origin` be changed to a `set[tag_origin]`?

lordwelch: > Should `tag_origin` be changed to a `set[tag_origin]`?

Let's leave it a list for now



> ```python
> assign_funcs = {
>             "overlay": self.assign_overlay,
>             "overwrite": self.assign_overwrite,
>             "overwrite_no_clear": self.assign_overwrite_no_clear,
>             "add_missing": self.assign_add_missing,
>             "combine_new": self.assign_combine_new,
>             "combine_current": self.assign_combine_current,
>         }
> ```

You're so close now replace these strings with the enums
`"overlay" -> OverlayMode.overlay`

lordwelch: And then the assign function you can change to
```python
assign = assign_funcs.get(mode, self.assign_overlay)
```

mizaki: With `tag_origin`; change it to a list and also add date/time to the `TagOrigin` class? Then it gives options for merging etc.

lordwelch: Adding a date doesn't make sense as there is nothing that would actually set it in a useable way. CT would set it and then it would save it and then it would disappear.

tests/genericmetadata_test.py
mizaki: Because of the commandline clear `-m`, copy can't be used but I'll get into that later. It should be clearer now in the code also.

tests/genericmetadata_test.py
mizaki: Now it should show how it's different with comes back to the commandline clear again

comicapi/genericmetadata.py
mizaki: I've cleaned this up now and use a `dict` to dedupe so it should be clearer.

tests/genericmetadata_test.py
mizaki: Again, because of the commandline clear?

mizaki:
I've think I finally see the issue with the `-m` and clearing list/sets. Using the wrong type for the commandline metadata for a list/set does seem the easiest solution. The other option might be: `Filename`, `commandline`, `read style(s)` and `talker` metadata have different `OverlayModes`. Maybe not all need to be a user available settings (but they could).

I've put comments of `Should always be "overlay" mode?` where I thought the option should be static.

Another (bad) idea would be to mark the origin of the data (maybe re-purpose/use TagOrigin) in the `GenericMetadata`.

tests/genericmetadata_test.py
lordwelch: All you've done is implement a type conversion as an overlay mode. Instead do the type conversion in the overlay method or do a manual check when the metadata from the `-m` option is applied

tests/genericmetadata_test.py
lordwelch: This doesn't show all of the types in play.
```python
def test_metadata_overlay_overwrite_no_clear():
    md = comicapi.genericmetadata.GenericMetadata(
        series="test", title="test", issue="1", volume_count=1, page_count=3, day=3, genres={"test", "test2"}, story_arcs=["arc"], characters={"bob", "fred"}
    )
    overwrite = comicapi.genericmetadata.GenericMetadata(
        series="test2", title="", issue_count=2, page_count=2, day=0, genres={'fake'}, characters="", scan_info="nothing",
    )
    expected = comicapi.genericmetadata.GenericMetadata(
        series="test2",
        title="test",
        issue="1",
        issue_count=2,
        volume_count=1,
        day=0,
        page_count=2,
        genres={"fake"},
        story_arcs=["arc"],
        characters={"bob", "fred"},
        scan_info="nothing",
    )
    md.overlay(overwrite, OverlayMode.overwrite_no_clear)

    assert md == expected
```

With this in place we test far more.

- If a `str` is set on both the second value is used: `series`
- If a `str` is set on the first and on the second an empty string is set the first is used: `title`
- If a `str` is set on the first and on the second it is __not__ set the first is used: `issue`
- If a `str` is __not__ set on the first and on the second it is set the second is used: `scan_info`
- If a `int` is set on the first and on the second it is __not__ set the first is used: `volume_count`
- If a `int` is set on both the second value is used: `page_count`
- If a `int` is __not__ set on the first and on the second it is set the second is used: `issue_count`
- If a `int` is set on the first and on the second a `0` is set the second is used: `day`
- If a `set` is set on both the second value is used: `genres`
- If a `list` is set on the first and on the second it is __not__ set the first is used: `story_arcs`
- If a `set` is set on the first and on the second an empty string is set the first is used: `characters`

Now that there is a more complete test, where would this be used? Unless I am mistaken all of these cases are taken care of elsewhere in comictagger. In taggerwindow.py `form_to_metadata` utils.xlate is used to ensure empty string values are set to `None` the same is done with numbers except for `critical_rating` which is special cased at the value 0.0 to be set to `None` because the control doesn't allow it being unset. The same is used in comicrack.py `_convert_xml_to_metadata` except for the special case for `critical_rating` (which means that if someone explicitly set the value to 0 ComicTagger cannot handle it properly and will unset it)

lordwelch:
You use the quirk on the commandline that an empty string value clears a list as a reason for why these options need to exist..... And then you explicitly say that none of the options you created should be used......
```python
        md.overlay(self.config.Runtime_Options__metadata)  # Should always be "overlay" mode?
```

mizaki:
> You use the quirk on the commandline that an empty string value clears a list as a reason for why these options need to exist..... And then you explicitly say that none of the options you created should be used......
> 
> ```python
>         md.overlay(self.config.Runtime_Options__metadata)  # Should always be "overlay" mode?
> ```

It depends on the answer to the question of keep using the quirk or not? If it is used then `OverlayMode.overwrite` for example is required for that reason (i.e. having to check if a list is a list). If, say, `TagOrigin` is used to demarcate if the data is from commandline, file name, etc. then `OverlayMode.overwrite` could be the same (and thus removed and do a copy). It could also be that
```python
# finally, use explicit stuff
md.overlay(self.config.Runtime_Options__metadata)
```
Does not use the user available setting and is always `OverlayMode.overlay` with the `list/set == str` in whereas the other modes do not check.

Having the user choose for all four of the origins (file name, read style, commandline and talker) might lead to confusion?

lordwelch:
> It depends on the answer to the question of keep using the quirk or not

No. You have several different modes that you have created. When I said to remove them or asked why they exist you explicitly said that they existed because of the quirk and then you explicitly don't use them the only place the quirk happens. Remove the useless modes. It's a stupid quirk that probably shouldn't ever have been added in the first place.

If something needs to change about how metadata is merged from different sources that needs to change where the decision is made, in this case cli.py `create_local_metadata`. Creating an entire complicated system to facilitate a quirk is stupid. It was implemented as a tiny 7 line change in 539aac1307937ed53ad851496fe5b0f288886a66 to allow unsetting the credits #587. If you can turn this into a 50 line change (minus tests) I'll merge it as is but this is over 700 lines changed, most of which is just to satisfy a stupid quirk that doesn't even get executed.

mizaki:
I've squished and targeted what is currently already there. The overlay options will only apply to source (talker) information currently. So this option will become more relevant when the multi-read of metadata styles is done.

I have also renamed `Comic Book Lover` to `Metadata Options` and moved those that are to do with metadata into it.

comicapi/genericmetadata.py
```diff
@@ -202,96 +250,190 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
-                        getattr(self, cur).clear()
-                    else:
-                        setattr(self, cur, None)
-                elif isinstance(new, (list, set)) and len(new) == 0:
-                    pass
-                else:
-                    setattr(self, cur, new)
+    def norm_fold(self, string: str) -> str:
+        """Normalise and casefold string"""
+        return unicodedata.normalize("NFKD", string).casefold()
```
lordwelch: Lets move this to utils

comicapi/genericmetadata.py
```diff
@@ -202,96 +250,190 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+
+        # Create dict values for deduplication
+        new_dict: dict[str, str] = {self.norm_fold(n): n for n in new}
+        cur_dict: dict[str, str] = {self.norm_fold(c): c for c in cur}
+
+        if isinstance(new, list) and isinstance(cur, list):
+            cur_dict.update(new_dict)
+            return list(cur_dict.values())
+
+        if isinstance(new, set) and isinstance(cur, set):
```
lordwelch: I don't think we need to do the isinstance check on `new'

comicapi/genericmetadata.py
```diff
@@ -202,96 +250,190 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
             if c["person"] == "":
-                for r in reversed(self.credits):
+                for r in reversed(cur_credits):
                     if r["role"].casefold() == c["role"].casefold():
-                        self.credits.remove(r)
+                        cur_credits.remove(r)
             # otherwise, add it!
             else:
-                self.add_credit(c["person"], c["role"], primary)
+                cur_credits.append(Credit(person=c["person"], role=c["role"], primary=primary))
```
lordwelch: All of the credit appends need to do the same checks as self.add_credit


It might be better to merge them and then clear the current credits and just add them at the end
```python
credits = assign_credit_func(cur_credits, new_credits)
self.credits.clear()
for credit in credits:
    self.add_credit(**credit)
```

comictaggerlib/ui/settingswindow.ui
```diff
@@ -401,17 +391,182 @@
...
+                 <width>16777215</width>
+                 <height>16777215</height>
+                </size>
+               </property>
+               <property name="toolTip">
+                <string>Overlay - When the new source metadata has non-empty values, overwrite the current metadata
+
+Add Missing - Any values that are empty in the current metadata that are in the new source metadata, add those values
+
+Combine - Combine lists and use the new (non-empty) source metadata for all other values</string>
```
lordwelch: I don't really like the wording here. What do you think of
 
> Overlay - Overlays new metadata on top of the original.
> (Series=batman, Issue=1, Tags=[batman,joker,robin])
> \+ (Series=Batman, Publisher=DC Comics, Tags=[mystery,action])
> = (Series=Batman, Issue=1, Publisher=DC Comics, Tags=[mystery,action])
> 
> Add Missing - Metadata missing from the original is added from the the new metadata.
> (Series=batman, Issue=1, Tags=[batman,joker,robin])
> \+ (Series=Superman, Issue=10, Publisher=DC Comics, Tags=[superman,lois lane,lex luthor])
> = (Series=batman, Issue=1, Publisher=DC Comics, Tags=[batman,joker,robin])
> 
> Combine - Same as Overlay except lists (tags, characters, genres, credits, etc...) are combined.
> (Series=batman, Issue=1, Tags=[batman,joker,robin])
> \+ (Series=Batman, Publisher=DC Comics, Tags=[mystery,action])
> = (Series=Batman, Issue=1, Publisher=DC Comics, Tags=[batman,joker,robin,mystery,action])

But now that's a lot more text....

How about making this a radio box so each item can have it's own description on the screen and we can make the example the tool tip?

mizaki: Examples should make things easier to understand.

I'm thinking of a text box (on the right as there is space) that changes to reflect the current chosen or, a tabbed text box for the options.

This is kind of related but I think it makes sense to have a choice for the read styles and the source data (with commandline always in `overlay` with clearing of lists etc. and having primacy over read styles). So having two combo boxes makes the most sense then rather than radio.

comicapi/genericmetadata.py
mizaki: I have kept it a return value so it's not different to the others. I've also removed the empty person to remove role.

As the `add_credit` is done to each `md` I did a dupe check so using `add_credit` isn't used again. I think it makes sense but if you want to still use it, I'll change it.

comictaggerlib/ui/settingswindow.ui
mizaki: This is what I went with. Could possible have it switch tab when a combo box is changed.
image

lordwelch: Read style and source info need to have similar text as they are the same operation but the metadata comes from a different place

comictaggerlib/ui/settingswindow.ui
```diff
@@ -450,99 +545,344 @@
...
+                <property name="rightMargin">
+                 <number>0</number>
+                </property>
+                <property name="bottomMargin">
+                 <number>0</number>
+                </property>
+                <item>
+                 <widget class="QPlainTextEdit" name="combineTextEdit">
+                  <property name="plainText">
+                   <string>Same as Overlay except lists (tags, characters, genres, credits, etc.) are combined.
```
lordwelch: This is not accurate with the current code anymore

comicapi/genericmetadata.py
```diff
@@ -339,19 +477,20 @@ def get_cover_page_index_list(self) -> list[int]:
...
-        for c in self.credits:
-            if c["person"].casefold() == person.casefold() and c["role"].casefold() == role.casefold():
-                # no need to add it. just adjust the "primary" flag as needed
-                c["primary"] = primary
-                found = True
-                break
-
-        if not found:
-            self.credits.append(credit)
+        if person != "":
```
lordwelch:
```suggestion
        if person == "":
            return
```

comictaggerlib/settingswindow.py
```diff
@@ -194,6 +194,9 @@ def __init__(
         )
         self.cbFilenameParser.clear()
         self.cbFilenameParser.addItems(utils.Parser)
+        for mem in OverlayMode:
```
lordwelch:
```suggestion
        for mode in OverlayMode:
```

comictaggerlib/taggerwindow.py
```diff
@@ -1166,7 +1166,8 @@ def query_online(self, autoselect: bool = False, literal: bool = False) -> None:
                             description=cleanup_html(
                                 new_metadata.description, self.config[0].Sources__remove_html_tables
                             ),
-                        )
+                        ),
+                        self.config[0].Metadata_Options__source_overlay,  # command line?
```
lordwelch: What's the question here?

comicapi/genericmetadata.py
```diff
@@ -204,96 +252,186 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+            return cur
+
+    def assign_combine(self, cur: Any, new: Any) -> Any:
+        """Combine - Combine lists and sets. All non-None values from new replace cur(rent)"""
+        if new is None:
+            return cur
+        if isinstance(new, (list, set)) and isinstance(cur, (list, set)):
+            if len(new) > 0 and isinstance(new, list) and isinstance(new[0], Url):
+                return self.weblinks_dedupe(new, cur)  # type: ignore[arg-type]
+            return self.assign_dedupe(new, cur)
```
lordwelch: This fails when there are no new links and there is a current link. Probably a good test case.

comicapi/genericmetadata.py
```diff
@@ -204,96 +252,186 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
+        if len(new) == 0:
+            return cur
+        if len(cur) == 0:
+            return new
+
+        new_set: set[str] = {n.url for n in new}
+        cur_set: set[str] = {c.url for c in cur}
+        new_set.update(cur_set)
+
+        return [parse_url(url) for url in new_set]
```
lordwelch: This is mostly unnecessary. The Url object is a tuple and the url property (actually a method) only does basic null checks and concatenation no sanitizing is done therefore this can be deduplicated with a simple set.
```python
list(set(new).union(cur))
```

comictaggerlib/ui/settingswindow.ui
mizaki: What about covering both? Something like:
```
Source Info: Overlays the new 'source info'  metadata on top of the current metadata as taken from the file name parser or read style (unless 'Overwrite metadata was selected)
Read Style: Overlays subsequent metadata after the first read style from the next read style as set by the user
```
The read style text above probably needs some simplification but that general idea?

comictaggerlib/ui/settingswindow.ui
mizaki: How's this:
```
Combines lists (tags, characters, genres, credits, etc.) replacing any duplicates with the new record. All other non-empty values replace any current.
```

comictaggerlib/taggerwindow.py
mizaki: That was a question I meant to ask: On the command line the styles from the `file` setting can be overridden but not saved. Should overlay do the same?

comictaggerlib/ui/settingswindow.ui
lordwelch: Is there a reason for the code to not match the description?

comictaggerlib/ui/settingswindow.ui
lordwelch: So we have the origin of the data "read", "source" and then we have "style" indicating there is some sort of style to the reading, then we have "info" which doesn't contribute any real meaning and doesn't follow the other control. They need to have consistent names. Imagine if on the front page we had this
Screenshot 2024-03-17 at 4 19 18 PM
That just adds confusion, because instead of keeping consistent names for the controls it uses completely different terms. I really care what terminology you use but it needs to be consistent. And if the name can indicate that "source" is from comicvine (or one of the others) then it needs a description as well (either as a tooltip or a label)

mizaki: IIRC you said to use "style" for the read and modify when it's user facing so I did that. "Data Source" IIRC I changed because of the space it takes up in the tagger window. And then in the settings it's still "Metadata Source", so I think that's the root of the problem I made. In the auto-tag window the option is labelled `Overwrite metadata` so maybe removing the word "Source" from the tagger window and simply put "Metadata" is the answer?

The tool tip for both the label and combox for source info is:
```
The operation to perform when overlaying source data (Comic Vine, Metron, GCD, etc.)
```
and for read style(s):
```
The operation to perform when overlaying read styles (ComicRack, ComicBookInfo, etc.)
```
Are they okay as is?

And the general format of using the one tab to explain both "Metadata" and "Read Style(s)" is okay?

comictaggerlib/ui/settingswindow.ui
mizaki: What am I missing?
```
Combines lists (tags, characters, genres, credits, etc.) replacing any duplicates with the new record. All other non-empty values replace any current.
```
I think that is correct? "replacing" maybe isn't the perfect word, "using the new value with any duplicate items"?

lordwelch: The current description says "Same as Overlay except for lists...." but, it isn't, can we just make the code match the description?

comictaggerlib/ui/settingswindow.ui
lordwelch: You're missing the point. That was just an example. I don't really care what terminology you use, it just needs to be consistent with the control it is next to.

comictaggerlib/ui/settingswindow.ui
mizaki: I'm with you now. It was a 50/50; change the description or change the code and I chose poorly.

If the command line empty string is taken out (presuming that command line metadata will always be used in `overlay` mode) then they look to do the same outside of lists/sets?
```python
    def assign_overlay(self, cur: Any, new: Any) -> Any:
        """Overlay - When the 'new' object has non-None values, overwrite 'cur'(rent) with 'new'."""
        if new is None:
            return cur
        # An empty string can be used to clear a list or set via the commandline
        if isinstance(new, str) and len(new) == 0:
            if isinstance(cur, (list, set)):
                cur.clear()
                return cur
            else:
                return None
        elif isinstance(new, (list, set)) and len(new) == 0:
            return cur
        else:
            return new
```
```python
    def assign_combine(self, cur: Any, new: Any) -> Any:
        """Combine - Combine lists and sets. All non-None values from new replace cur(rent)"""
        if new is None:
            return cur
        if isinstance(new, (list, set)) and isinstance(cur, (list, set)):
            if len(new) > 0 and isinstance(new, list) and isinstance(new[0], Url):
                return list(set(new).union(cur))
            return self.assign_dedupe(new, cur)
        else:
            return new
```
Apologies if I'm missing the obvious.

comictaggerlib/ui/settingswindow.ui
mizaki: I've changed it to "Data Source" to match the tagger window.

I've changed the description to this:
```
Read Style: Overlays all the none empty values of the read style metadata on top of the current metadata (i.e. file name parsing, other read style(s)).
Data Source: Overlays all the none empty values from the data source (e.g. Comic Vine) on top of the current metadata (i.e. file name parsing, read style(s)).
```
I think now that with the wording there is no need for separate descriptions for "Read Style" and "Data Source"?

mizaki:
I've rebased and squished.

I've gone back through and I don't think there are any outstanding (or accidental reversions).

Also changed the descriptions in the setting window which I hope covers it all now.

I've moved the GUI overlay options to the internal namespace and added CLI version that will override in the same way as `--types` does.

An "add missing with combine (lists/sets)" might be another option to add.

The credits might be better as you suggested as a combine current and new and then pipe them all through `add_credit`. The only reasons not to is it's maybe easier to understand as is, switching of if new overwrites or not (but that could be worked around), and then finally the possibility of adding another overlay mode for credits. Although, I have no idea what that might be right now.

And then there is the question of the `tag_origin`. I imagine you'd want any changes to handling the `notes` field to be a separate PR? So, leave it as is? Maybe the read styles could use it as well? Thinking out loud there...

lordwelch:
I rebased the merges out so it's only your new commits

mizaki:
> I rebased the merges out so it's only your new commits

Sorry, I usually check but obviously didn't.

comictaggerlib/cli.py
```diff
@@ -265,12 +268,12 @@ def create_local_metadata(self, ca: ComicArchive) -> GenericMetadata:
...
                     t_md = ca.read_metadata(style)
-                    md.overlay(t_md)
+                    md.overlay(t_md, self.config.internal__load_data_overlay)
                     break
                 except Exception as e:
                     logger.error("Failed to load metadata for %s: %s", ca.path, e)
 
-        # finally, use explicit stuff
+        # finally, use explicit stuff (always 'overlay' mode)
         md.overlay(self.config.Runtime_Options__metadata)
```
lordwelch: Set the mode argument explicitly here `mode=comicapi.genericmetadata.OverlayMode.overlay`

comicapi/genericmetadata.py
```diff
@@ -354,12 +472,15 @@ def get_cover_page_index_list(self) -> list[int]:
...
+        if person == "":
+            return
+
         credit = Credit(person=person, role=role, primary=primary)
 
         # look to see if it's not already there...
         found = False
         for c in self.credits:
-            if c["person"].casefold() == person.casefold() and c["role"].casefold() == role.casefold():
+            if norm_fold(c["person"]) == norm_fold(person) and norm_fold(c["role"]) == norm_fold(role):
```
lordwelch: The norm_fold on `person` and `role` should happen outside the loop

comicapi/genericmetadata.py
```diff
@@ -214,101 +261,172 @@ def get_clean_metadata(self, *attributes: str) -> GenericMetadata:
...
-                elif isinstance(new, (list, set)) and len(new) == 0:
-                    pass
+    def credit_dedupe(self, cur: list[Credit], new: list[Credit]) -> list[Credit]:
+        if len(new) == 0:
+            return cur
+        if len(cur) == 0:
+            return new
+
+        # Create dict for deduplication
+        new_dict: dict[str, Credit] = {f"{norm_fold(n['person'])}_{n['role'].casefold()}": n for n in new}
```
lordwelch: norm_fold the entire string instead of individually

lordwelch (changes requested):
I still want to look at the settings and what the tests cover but otherwise this should be just about done

comictaggerlib/cli.py
mizaki: I tried `mode=comicapi.genericmetadata.OverlayMode.overlay` but `comicapi` isn't imported so I imported `OverlayMode`.

mizaki:
> I still want to look at the settings and what the tests cover but otherwise this should be just about done

I've moved some of the data into `testing/comicdata.py`. I wasn't sure how to do the credits so left it as a possible "tomorrow" job. Some data might be able to be combined to make it less piecemeal.

mizaki:
Only other possible change is `tag_origin` to a list?

mizaki:
I've added the "lists" option. I feel like something is missing now but I have no idea what.

Two dropdowns with only two options seems too few!

mizaki:
Thinking about "lists", perhaps it should be three options: merge, replace or ignore (leave original data)?

lordwelch:
replace is the wrong word. It falls back to whatever the merge mode is

lordwelch (approved):

mizaki:
"ignore" might be the final option needed for attributes and lists. That way a user can update only attributes or lists. That would bring lists to a select the same as attributes and bring attributes to three options as well.

Shall I add an ignore and convert "lists" options from a `bool`?

lordwelch:
I don't see why someone would treat lists like that. Let's wait until someone has a usecase for it

PR #599, opened 2024-02-11
Implement local plugins

lordwelch:
Implements loading local plugins from the config dir.

For plugins that don't need external dependencies eg gcd-talker they can be cloned directly into the plugins directory

```bash
# rm -rf ./temp # careful with this command
comictagger --config temp --list-plugins
mkdir -p temp/plugins
cd temp/plugins
git clone https://github.com/comictagger/gcd_talker
cd -
comictagger --config temp --list-plugins
```

### DIFF
--- a/comicapi/comicarchive.py
+++ b/comicapi/comicarchive.py
@@ -16,12 +16,15 @@
 from __future__ import annotations
 
 import io
+import itertools
 import logging
 import os
 import pathlib
 import shutil
 import sys
 import traceback
+from collections.abc import Sequence
+from typing import TYPE_CHECKING
 
 from comicapi import utils
 from comicapi.archivers import Archiver, UnknownArchiver, ZipArchiver
@@ -29,32 +32,37 @@ from comicapi.genericmetadata import GenericMetadata
 from comicapi.metadata import Metadata
 from comictaggerlib.ctversion import version
 
+if TYPE_CHECKING:
+    from importlib.metadata import EntryPoint
+
 logger = logging.getLogger(__name__)
 
 archivers: list[type[Archiver]] = []
 metadata_styles: dict[str, Metadata] = {}
 
 
-def load_archive_plugins() -> None:
+def load_archive_plugins(local_plugins: Sequence[EntryPoint] = tuple()) -> None:
     if not archivers:
         if sys.version_info < (3, 10):
             from importlib_metadata import entry_points
         else:
             from importlib.metadata import entry_points
         builtin: list[type[Archiver]] = []
-        for arch in entry_points(group="comicapi.archiver"):
+        # A list is used first matching plugin wins
+        for ep in itertools.chain(local_plugins, entry_points(group="comicapi.archiver")):
+            logger.warning(ep)
             try:
-                archiver: type[Archiver] = arch.load()
-                if arch.module.startswith("comicapi"):
+                archiver: type[Archiver] = ep.load()
+                if ep.module.startswith("comicapi"):
                     builtin.append(archiver)
                 else:
                     archivers.append(archiver)
             except Exception:
-                logger.exception("Failed to load archive plugin: %s", arch.name)
+                logger.exception("Failed to load archive plugin: %s", ep.name)
         archivers.extend(builtin)
 
 
-def load_metadata_plugins(version: str = f"ComicAPI/{version}") -> None:
+def load_metadata_plugins(version: str = f"ComicAPI/{version}", local_plugins: Sequence[EntryPoint] = tuple()) -> None:
     if not metadata_styles:
         if sys.version_info < (3, 10):
             from importlib_metadata import entry_points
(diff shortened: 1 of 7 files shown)


comicapi/comicarchive.py
```diff
@@ -16,68 +16,78 @@
...
     if not archivers:
         if sys.version_info < (3, 10):
             from importlib_metadata import entry_points
         else:
             from importlib.metadata import entry_points
         builtin: list[type[Archiver]] = []
-        for arch in entry_points(group="comicapi.archiver"):
+        # A list is used first matching plugin wins
+        for ep in itertools.chain(local_plugins, entry_points(group="comicapi.archiver")):
+            logger.warning(ep)
```
mizaki: Left in for testing?

comicapi/comicarchive.py
```diff
@@ -16,68 +16,78 @@
...
         if sys.version_info < (3, 10):
             from importlib_metadata import entry_points
         else:
             from importlib.metadata import entry_points
         builtin: dict[str, Metadata] = {}
         styles: dict[str, Metadata] = {}
-        for arch in entry_points(group="comicapi.metadata"):
+        # A dict is used, last plugin wins
+        for ep in itertools.chain(entry_points(group="comicapi.metadata"), local_plugins):
+            logger.warning(ep)
```
mizaki: Testing?

comictaggerlib/main.py
```diff
@@ -123,9 +123,20 @@ def run(self) -> None:
         self.main()
 
     def load_plugins(self, opts: argparse.Namespace) -> None:
-        comicapi.comicarchive.load_archive_plugins()
-        comicapi.comicarchive.load_metadata_plugins(version=version)
-        self.talkers = comictalker.get_talkers(version, opts.config.user_cache_dir)
+        local_plugins = plugin_finder.find_plugins(opts.config.user_plugin_dir)
+        self._extend_plugin_paths(local_plugins)
+        logger.warning(sys.path)
```
mizaki: Testing?

comictaggerlib/ctsettings/types.py
```diff
@@ -49,6 +49,13 @@ def user_log_dir(self) -> pathlib.Path:
             return path
         return pathlib.Path(super().user_log_dir)
 
+    @property
+    def user_plugin_dir(self) -> pathlib.Path:
+        if self.path:
+            path = self.path / "plugins"
+            return path
+        return pathlib.Path(super().user_config_dir)
```
mizaki, line 57: Is the `present working directory` expected to always be the root of comictagger? It may have been my scuffed setup but I had `comictaggerlib` as my `pwd` and so it was looking for `./comictagger/comictaggerlib/temp/plugins`

comictaggerlib/ctsettings/plugin_finder.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def find_plugins(plugin_folder: pathlib.Path) -> Plugins:
+    """Discovers all plugins (but does not load them)."""
+    ret: list[Plugin] = []
+    for plugin_path in plugin_folder.glob("*/setup.cfg"):
+        try:
+            ret.extend(_find_local_plugins(plugin_path.parent))
+        except Exception as err:
+            FailedToLoadPlugin(plugin_path.parent.name, err)
```
mizaki, line 126: Is the `present working directory` expected to always be the root of comictagger? It may have been my scuffed setup but I had `comictaggerlib` as my `pwd` and so it was looking for `./comictagger/comictaggerlib/temp/plugins`

comictalker/__init__.py
```diff
@@ -21,12 +23,15 @@
...
+) -> dict[str, ComicTalker]:
     """Returns all comic talker instances"""
     talkers: dict[str, ComicTalker] = {}
     ct_version = parse(version)
 
-    for talker in entry_points(group="comictagger.talker"):
+    # A dict is used, last plugin wins
+    for talker in itertools.chain(entry_points(group="comictagger.talker"), local_plugins):
         try:
             talker_cls = talker.load()
```
mizaki, line 36: Going by this I would expect the plugin in the `plugins` dir to be the last and running through the loop it is but, the path from my test is the same as the first plugin. That is: I have `gcd` installed via pip and if I also put the `gcd` plugin in the `plugins` dir (with a change to the `about` text mark the difference), I can see `plugin_finder` finds the `gcd` in the `plugins` dir with the correct path but when it comes here later and does the `talker.load()` it uses the (or hasn't changed anything) dir from the first `gcd` found (the installed via pip). Hope that makes sense?

mizaki:
I think it should be logged which plugin (and dir) was used in case someone uses a custom (and broken) version by overriding the default. Otherwise, there would be no way to tell from the logs alone?

comictalker/__init__.py
lordwelch: They have the same import path, so it looks through sys.path and finds the installed one first

comicapi/comicarchive.py
lordwelch: Yeah...

comictaggerlib/ctsettings/plugin_finder.py
lordwelch: I'll have to check glob but it should start from plugin_folder

lordwelch: That is because your command had a local path eg `comictagger --config temp` if you had `comictagger --config /path/to/comictagger/temp` it would have found it correctly. If you look in `./comictagger/comictaggerlib/temp/` you should see the config, cache and logs as well

comictalker/__init__.py
lordwelch: So it might be a good idea in the future to load something better but for now they will have to use distinct import paths. So for gcd to override it you would need to used a different directory name and update the entrypoint.
Dependencies can be included by extracting the wheel into the plugin directory eg `plugins/gcd_talker/mokkari` would allow `import mokkari` to work. Note that bundled modules will take priority so on any real release the bundled mokkari would be used instead of the mokkari in a plugin folder

lordwelch:
> I think it should be logged which plugin (and dir) was used in case someone uses a custom (and broken) version by overriding the default. Otherwise, there would be no way to tell from the logs alone?

Done

comictalker/__init__.py
mizaki: The deps thing is something to put in the wiki, I will see about rounding it out now we're near a freeze.

mizaki:
Seems good from my testing. It loaded a copy of GCD plugin `gcd2` and it worked. I tried importing a non-installed module and it was all logged.

PR #601, opened 2024-02-17
Support multiple web_links

lordwelch:


lordwelch:
It still needs a better control than a single line edit.
This is implemented to include more compatibility with ComicRack see https://github.com/anansi-project/comicinfo/issues/56

mizaki:
The web link in the GUI remains red.
comictagger.py:115
```python
"web_link": self.leWebLink,
```
`web_link` should be `web_links`

mizaki:
The web link button doesn't work correctly, send the whole line with both URLs to browser.

Could change the web link line edit to a combo box and then clicking the visit button goes to whatever's picked but it could be confusing. Users might think it's the URL they want to save.

A multiline edit with the URLs themselves as normal links, ideally it would expand vertically for each line. This is the best bad idea, I think.

mizaki:
This is what I see when I go to the `Notes` tab:
image
There is no highlighted line (it makes sense it's is the top line though), maybe helpful to highlight the top line.

Similar when adding a line:
ct_weblinks
Here CT could add `https://` and set the line as editing after the `add` button is pressed.

With other themes all this may be more obvious?

comictaggerlib/taggerwindow.py
```diff
@@ -550,6 +550,7 @@ def add_weblink_item(self, url: str = "") -> None:
             | QtCore.Qt.ItemFlag.ItemIsDragEnabled
             | QtCore.Qt.ItemFlag.ItemIsSelectable
         )
+        self.leWebLink.item(self.leWebLink.count() - 1).setSelected(True)
```
mizaki, line 553:
```python
        if url:
            self.leWebLink.item(self.leWebLink.count() - 1).setSelected(True)
        else:
            # url will be False from the `Add Item` button
            self.leWebLink.editItem(self.leWebLink.item(self.leWebLink.count() - 1))
```
How about that? `url` is `False` when the button is pressed.

mizaki:
Is it just me? I can no longer edit the fields and the text is "greyed out"? I don't seem to be able to interact with it at all, even the scroll bar doesn't work.
web_links2

lordwelch:
Are you editing comicrack metadata?

mizaki:
> Are you editing comicrack metadata?

Ah, it was Comic Info XML which I thought would be the same as Comic Rack.

It does show that (stupid) people might need the buttons disabled too.

Other than that, it looks good.

lordwelch:
Yeah cix doesn't have the attribute name updated yet because it's in a different repo

PR #610, opened 2024-02-24
ii restructure

lordwelch:
Restructures the IssueIdentifier into more manageable sections.
Only minimal changes are implemented here this is not meant to change the results this is just meant to be a restructure.


comictaggerlib/issueidentifier.py
```diff
@@ -543,142 +493,188 @@ def search(self) -> list[IssueResult]:
...
+                    cover_matching_2.append(score)
+
+            if len(cover_matching_2) == 0:
+                if len(cover_matching_1) == 1:
                     self.log_msg("No matching pages in the issue.")
                     self.log_msg("--------------------------------------------------------------------------")
-                    print_match(self.match_list[0])
+                    self._print_match(cover_matching_1[0])
                     self.log_msg("--------------------------------------------------------------------------")
                     self.search_result = self.result_found_match_but_bad_cover_score
```
mizaki: This returns to alter `self.search_result` but there are two problems. L:660 would overwrite it if it didn't use `search_result` instead. So a low confidence result becomes a good one.

lordwelch: Fixed

mizaki:
I've not found any other issues.

PR #616, opened 2024-02-27
UI/Text Improvements

lordwelch:


comictaggerlib/ctsettings/file.py
```diff
@@ -41,223 +40,292 @@ def internal(parser: settngs.Manager) -> None:
...
-        default=False,
+    parser.add_setting("--assume-lone-credit-is-primary", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-characters-to-tags", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-teams-to-tags", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-locations-to-tags", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-storyarcs-to-tags", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-notes-to-comments", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--copy-weblink-to-comments", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--apply-transform-on-import", default=False, action=argparse.BooleanOptionalAction)
+    parser.add_setting("--apply-transform-on-bulk-operation", default=False, action=argparse.BooleanOptionalAction)
```
mizaki, line 155: In the GUI there is the group title `CBL Options`/`CBL Transforms` but removing the `cbl` on the cli makes it unclear these relate only to CBL, as say CIX has tags too etc. Maybe a simple `help="Relates to CBL only"` or similar?

mizaki:
image
`Clear all existing tags during metadata download, default is to merge downloaded metadata with existing tags.` is a little unwieldy, I think I chopped it off (`Clear all existing tags during metadata download`) in one of my PRs and added the full thing in the tool tip.

lordwelch:
> image
> `Clear all existing tags during metadata download, default is to merge downloaded metadata with existing tags.` is a little unwieldy, I think I chopped it off (`Clear all existing tags during metadata download`) in one of my PRs and added the full thing in the tool tip.

I edited the xml by hand on that one

comictaggerlib/ctsettings/file.py
lordwelch: I'll have to double check but I do not believe it's ever actually restricted to the CBL type. Also those options are long enough as is

mizaki: I believe it is only CBL, random example from `cli.py:L353` `if self.config.Metadata_Options__apply_transform_on_bulk_operation and dst_tag_id == "cbi":`?
Although, taking the `dst_tag_id == "cbi"` out is the only reason I see.

If it's changed to all, the GUI should be changed:
image

lordwelch: That's for copying tags, all of the download/auto tag operations use prepare_metadata. I believe it's always been that way, but I've been wrong before and I'm not at a computer to check.

In any case the GUI needs at most a clarifying tooltip

lordwelch: Even back when it was all loose files in the root directory the auto-tag would always apply it to all tags.
I added a tool tip explaining that it applies to all tags
Screenshot 2024-06-21 at 7 09 14 PM

Issue #261, opened 2022-04-07
Anansi Project: Schema v2.1: StoryArcNumber


MichaelFitzurka:
Like the Number field, but for StoryArcs.

We can add this to the Details tab, where the Volume field is now (to align with Story Arc) 
and then shift the other left hand values down one.

MichaelFitzurka:
I have the feature already completed - but it was mentioned in another ticket that you were looking for a way to separate out 2.1 from 2.0 support.
Awaiting that plugin/modification.

nicolasshu:
Hey! Are we still looking to get this feature implemented? I was working on this feature then I saw that you've already implemented

lordwelch:
ComicInfoXml support for the anansi project https://github.com/anansi-project/comicinfo lives here https://github.com/comictagger/comicinfoxml.

Pull requests are welcome to add GUI support for the Story Arc Number property, but it should provide a useful way to correlate multiple values to the Story Arc property see https://anansi-project.github.io/docs/comicinfo/documentation#storyarcnumber for more information

tehguitarist:
@MichaelFitzurka or @nicolasshu, I know the codebase has changed quite a bit since you first wrote the implementation, is this something you'd be willing to update and merge?

nicolasshu:
I have a lot of stuff going on right now, so I won't have time in the next couple of weeks, but after that, I might be able to

tehguitarist:
I'll see if I can draft something up, but it'll likely need a good review. I'll do my best to get things started.

tehguitarist:
Created initial PR's but will definitely need someone more experienced than me to have a look at them. Tests passed, but I couldn't build due to some issues I suspect is from me being on an ARM mac (I did try using the x86 python and exporting variables).

PR #622, opened 2024-03-10
[Feature] Add StoryArcNumber to main app.

tehguitarist:
Ok so this one definitely needs a review. 

I'm having issues running the build. I'm on a m1 Mac, and followed the build instructions in the contributing doc, but I'm very much an amateur here. Changes seem pretty minor, so hopefully I got it right.


lordwelch:
Yeah, I do a lot of my development on an M1 and there isn't a great way that I've found to build the pyinstaller package.

As I said in https://github.com/comictagger/comictagger/issues/261#issuecomment-1905152210 there needs to be a way to correlate this with the story arc title, right now there isn't a great way to see which number corresponds to which title.
Editing like this is very error prone and you have to go and count commas in order to figure out that the second story title doesn't even have a number, and it's incredibly easy to add other numbers that don't go along with a story arc.
Screenshot_20240310_121738

tehguitarist:
Yeah I was thinking about that. It's probably a bit out of my skills to refactor (not to mention the difficulty in making builds to test), but would a 2d array work? Storing each element as the arc and number? In terms of displaying, I suppose you could have distinct text boxes above and below, but that would necessitate a somewhat arbitrary limit, or the ability to dynamically add a pair of them.

I suppose the other way would be to append the number to the name, which could create issues if there is a story arc that ends in a number.

That being said, whilst magic numbers are never a good thing, I suspect we might struggle to find any individual issue that encompasses more then 3-5 story arcs. Maybe an author's run collected in HC though?

Perhaps a drop down that can be added to, which changes the value of the Story Arc Number based on the selection?

Whilst I might have the desire the see this through to completion, I might need to let someone who isn't a dabbler take it from here - I'd rather this get done right.

lordwelch:
I need to go over the CONTRIBUTING document again but after you activate the venv you can run `pip install pyqt5` to install gui support and then you can run `comictagger --config temp` to run the code you've changed

> I suppose the other way would be to append the number to the name, which could create issues if there is a story arc that ends in a number.

String parsing is almost always a bad idea, that's why I changed everything on the 'Other' tab to newline separated values as it's really hard for a character name to have a newline in it's name.

We just added a similar control for the weblinks in #601 you can get a pre-built download here to test https://github.com/comictagger/comictagger/actions/runs/8219353502 [download instructions]

The python side is fairly simple we can use a NamedTuple to represent a single story arc

```python
from typing_extensions import NamedTuple, Required

class StoryArc(NamedTuple):
    name: str
    number: str  # or maybe an `int | None` to force people to use it properly
    
class GenericMetadata:
    ...
    story_arcs: list[StoryArc]
```

Then the issue becomes: What sort of interface do we present to the user so that they can edit it?
Using a QListWidget like the weblinks probably isn't going to work out because we need to have two values, the name and the number, but a QTableWidget with two columns per row, or two rows per column might work.


[download instructions]: https://github.com/comictagger/comictagger/wiki/downloading-packages-from-pull-requests

tehguitarist:
I'll see what I can hack together, but this might be better tackled by someone with some stronger skills than mine. I think I could probably work out how to add some of the table widgets, but integrating and converting all the data touch points is probably a bit beyond my skills.

I think the cleanest look might to still have an add/remove button, which looks doable looking at the QT docs, but I'm unsure how to get the talkers to do that automatically when scraping, rather than just manual entry.

Not for lack of wanting to help, just knowing my limitations.

lordwelch:
For CV you would have to pull the story arc and just enumerate the index in the list https://comicvine.gamespot.com/api/documentation#toc-0-31
but that doesn't need to happen in this pr, you can just leave the number blank for now.

PR #627, opened 2024-04-06
Improve commandline metadata override

lordwelch:
Change parse_metadata_from_string to yaml syntax
Add a special value to remove existing values when metadata is overlayed


mizaki:
Looks good. The command line help needs to be updated though.

mizaki:
Also, as something for later, if `-m` (or a new one) could accept a `.yaml` file. (It could done with pipes etc. but just for ease.)

lordwelch:
I'll give it the `@file` syntax

PR #629, opened 2024-04-11
Multi read style

mizaki:
There's a bit to explain...

I tried a few ways to get the dropdown (delegate "widget" built using other widgets). I went with the `QTableWidget` as I think the headers (the order/# at least) is needed and so a table view was the easiest there. The `model` could be a list view but as the `QTableWidget` comes with a table model it was easier than having to create a new model and a `QTableView`.

Currently the selection is stored as a `dict[str, int]`, I'm not sure on the best way to go: A `list` could work using the `index` for the order. A `list` of `tuples` was another option I considered. I went with the `dict[str, int]` for the JSON saving (although the other options didn't seem to be much of a problem with JSON either).

I went with the custom `QTableWidgetItem` for sorting as using a `QSortFilterProxyModel` because it seems the `QTableWidget` won't allow a proxy model. It may be better to split the data of the `dict` between the `order/#` column (data: -1) and the style name (data: "cr") on the `read style` column. Then the display and data should make sorting easier.

The `QProxyStyle` to centre the check boxes had to be done because it was annoying me... :) Still need to change the background of the table to match the other `QComboBox`.

I went with sub-classing the standard `QComboBox` as the quite extensive changes didn't seem to make sense to subclass `CheckableComboBox`

The `QButtons` for moving the `read style` up and down seemed like the best solution (although I'm sure it's probably bad UX). I plan to add to the settings window a way to set the read and modify styles too.

The command line `-t --ttype` help message should populate the styles from `metadata_styles` similar to the `metavar`?

There are `TODO` questions which should (hopefully) make sense.

I expect I've forgotten some things too but I'm sure there's still plenty wrong I've missed.


comictaggerlib/ui/customwidgets.py
```diff
@@ -112,3 +114,384 @@ def toggleItem(self, index: int) -> None:
...
+
+
+class HoverQLabel(QtWidgets.QLabel):
+    """A QLabel with two QButtons that appear on hover"""
+
+    def __init__(self, text: str, parent: TableComboBox):
+        super().__init__(text, parent=parent)
+        self.combobox = parent
+
+        self.button_up = QtWidgets.QPushButton(QtGui.QIcon(str(graphics_path / "up.png")), "", self)
```
lordwelch: These aren't in the repo

comictaggerlib/ctsettings/file.py
```diff
@@ -25,7 +25,7 @@ def internal(parser: settngs.Manager) -> None:
     # automatic settings
     parser.add_setting("install_id", default=uuid.uuid4().hex, cmdline=False)
     parser.add_setting("save_data_style", default=["cbi"], cmdline=False)
-    parser.add_setting("load_data_style", default="cbi", cmdline=False)
+    parser.add_setting("load_data_style", default=[{"cbi": 0}], cmdline=False)
```
lordwelch: Wrong type

lordwelch:
> I tried a few ways to get the dropdown (delegate "widget" built using other widgets). I went with the `QTableWidget` as I think the headers (the order/# at least) is needed and so a table view was the easiest there. The `model` could be a list view but as the `QTableWidget` comes with a table model it was easier than having to create a new model and a `QTableView`.

Have you tried using a item delegate? https://doc.qt.io/qt-5/qitemdelegate.html

> Currently the selection is stored as a `dict[str, int]`, I'm not sure on the best way to go: A `list` could work using the `index` for the order. A `list` of `tuples` was another option I considered. I went with the `dict[str, int]` for the JSON saving (although the other options didn't seem to be much of a problem with JSON either).

Going with a list will be better and currently things can end up out-of order as the dict is never sorted.

> I went with the custom `QTableWidgetItem` for sorting as using a `QSortFilterProxyModel` because it seems the `QTableWidget` won't allow a proxy model. It may be better to split the data of the `dict` between the `order/#` column (data: -1) and the style name (data: "cr") on the `read style` column. Then the display and data should make sorting easier.

I'm not sure any sorting is really needed. I don't like that they re-sort whenever I enable one. Also since merging happens as an overlay the "top" item should be the last in the list when it is overlayed as the last item overlayed is what takes priority.

> The `QProxyStyle` to centre the check boxes had to be done because it was annoying me... :) Still need to change the background of the table to match the other `QComboBox`.

Most good things happen because someone is annoyed

> The `QButtons` for moving the `read style` up and down seemed like the best solution (although I'm sure it's probably bad UX). I plan to add to the settings window a way to set the read and modify styles too.


> The command line `-t --ttype` help message should populate the styles from `metadata_styles` similar to the `metavar`?

I have an updated message in #616

Also can you rebase this on the current develop branch

lordwelch:
And one of your todo's was to make a separate load and save arguments for the cli, yes that will be needed to merge this

mizaki:
> 
> Have you tried using a item delegate? https://doc.qt.io/qt-5/qitemdelegate.html
> 
I'll give it another look. I did try using `QStyledItemDelegate`. It was really the header that made me switch to `QTableWidget` but with the extra knowledge I have no I might find it better.
> 
> Going with a list will be better and currently things can end up out-of order as the dict is never sorted.
> 
Because the dict has the `order` it doesn't need to be sorted as the view sorts. It's only re-ordered `_setOrderNumbers` when an item is unchecked. Otherwise, `_move_item` alters the dict values. Anyway, using a list for the JSON should be easy enough. For the `UserRole` data, depending on the sorting, I'm thinking that it might need to store the `order` number as the `list` index might not be correct. Would a tuple be better in that case? `("cr", 1)`
> 
> I'm not sure any sorting is really needed. I don't like that they re-sort whenever I enable one. Also since merging happens as an overlay the "top" item should be the last in the list when it is overlayed as the last item overlayed is what takes priority.

The re-sort made the most sense as they are numbered. Would you rather they stayed in the same location and it's the `#/order/priority` label changes only? E.g.
```
- {unchecked} CoMet
2 {checked} Comic Rack
- {unchecked} XML
1 {checked} ComicBookInfo
```
```
1 {checked} CoMet
3 {checked} Comic Rack
4 {checked} XML
2 {checked} ComicBookInfo
```
I went back and forth on the ordering and reasoned that the average user would expect `1` to be the top priority. Naming the column "Priority" (although that takes up extra space) might help? It is "overlay-ed" but I don't think the average user would expect the data in "3" to overwrite the data in "1" etc. Or was it the visual order and not the numbering: "1, 2, 3" (top to bottom) but it should be displayed "3, 2, 1" (top to bottom)?

> Also can you rebase this on the current develop branch

Yep. I'll do the same for the `OverlayModes` too.

mizaki:
> And one of your todo's was to make a separate load and save arguments for the cli, yes that will be needed to merge this

`--type-modify` and `--type-read`? Is there any single letters that would make sense? 

Should the naming between the CLI and GUI be unified as well? "type" vs "style"

comictaggerlib/autotagmatchwindow.py
```diff
@@ -258,4 +266,4 @@ def save_match(self) -> None:
                 )
                 break
 
-        ca.load_cache(list(metadata_styles))
+        ca.load_cache(list(metadata_styles))  # TODO Should this be what the others do in taggerwindow.py etc.?
```
mizaki: `ca.load_cache([*self.load_data_styles, *self.save_data_styles])` is done in `taggerwindow.py`. Should this be only the read styles or all the styles?

comictaggerlib/cli.py
```diff
@@ -379,7 +378,9 @@ def copy(self, ca: ComicArchive) -> Result:
         except Exception as e:
             logger.error("Failed to load metadata for %s: %s", ca.path, e)
             return res
-        for style in self.config.Runtime_Options__type:
+        for style in self.config.Runtime_Options__type_modify:
+            if style == src_style_name:
+                continue
```
mizaki, line 371: Skipping writing same read style tags to same modify style tags.

comictaggerlib/ctsettings/commandline.py
```diff
@@ -160,14 +160,20 @@ def register_runtime(parser: settngs.Manager) -> None:
...
+        help="""Specify the type of tags to write.\nUse commas for multiple types.\nSee --list-plugins for the available types.\n\n""",
+        file=False,
+    )
+    parser.add_setting(
+        "--type-read",
         metavar=f"{{{','.join(metadata_styles).upper()}}}",
         default=[],
         type=metadata_type,
-        help="""Specify TYPE as either CR, CBL or COMET\n(as either ComicRack, ComicBookLover,\nor CoMet style tags, respectively).\nUse commas for multiple types.\nFor searching the metadata will use the first listed:\neg '-t cbl,cr' with no CBL tags, CR will be used if they exist\n\n""",
+        help="""Specify the type of tags to read.\nUse commas for multiple types.\nSee --list-plugins for the available types.\nThe tag use will be 'overlayed' in order:\ne.g. '-t cbl,cr' with no CBL tags, CR will be used if they exist and CR will overwrite any shared CBL tags.\n\n""",
```
mizaki, line 177: Will need to be altered slightly possibly when `overlayMode` goes in. Some comment like: "Check wiki for detailed example" could be added (with created wiki page).

comictaggerlib/renamewindow.py
```diff
@@ -81,7 +83,14 @@ def config_renamer(self, ca: ComicArchive, md: GenericMetadata | None = None) ->
...
         if md is None:
-            md = ca.read_metadata(self.data_style)
+            # TODO Same is used for taggerwindow.py:~1734/autotagmatchwindow Make a method?
+            md = GenericMetadata()
+            try:
+                for style in self.data_styles:
+                    md.overlay(ca.read_metadata(style))
+            except Exception as e:
+                logger.error("Failed to load metadata for %s: %s", ca.path, e)
+
```
mizaki: This bit of code is repeated in the mentioned windows. Is it worth getting them to use the same method?

comictaggerlib/cli.py
lordwelch: But should we? We now read all of them and combine them meaning that the series or issue number and other tags may have changed

comictaggerlib/taggerwindow.py
```diff
@@ -1219,24 +1221,25 @@ def commit_metadata(self) -> None:
...
-            self.config[0].internal__load_data_style = self.load_data_style
+            self.load_data_styles = self.cbLoadDataStyle.currentData()
+            self.config[0].internal__load_data_style = self.load_data_styles
             self.update_menus()
             if self.comic_archive is not None:
                 self.load_archive(self.comic_archive)
-        else:
-            self.cbLoadDataStyle.currentIndexChanged.disconnect(self.set_load_data_style)
-            self.adjust_load_style_combo()
-            self.cbLoadDataStyle.currentIndexChanged.connect(self.set_load_data_style)
```
mizaki: Is the cancel option supposed to work as I've found: "Cancel will keep the data in the form AND change the style."? If so, making the message clearer might be an idea.

comictaggerlib/taggerwindow.py
```diff
@@ -1408,28 +1411,39 @@ def adjust_source_combo(self) -> None:
...
+            self.cbLoadDataStyle.setItemChecked(self.cbLoadDataStyle.findData(style), False)
 
     def adjust_save_style_combo(self) -> None:
         # select the current style
         unchecked = set(metadata_styles.keys()) - set(self.save_data_styles)
         for style in self.save_data_styles:
-            self.cbSaveDataStyle.setItemChecked(self.cbLoadDataStyle.findData(style), True)
+            self.cbSaveDataStyle.setItemChecked(
+                self.cbSaveDataStyle.findData(style), True
+            )  # Why were these read style?
```
mizaki: I think I went back in the history and it was always `self.cbLoadDataStyle.findData(style)`. It's going to be the same `cr` etc. but seemed odd. Possibly a copy/paste?

comictaggerlib/taggerwindow.py
```diff
@@ -1620,7 +1634,8 @@ def copy_tags(self) -> None:
         ca_list = self.fileSelectionList.get_selected_archive_list()
         has_src_count = 0
 
-        src_style = self.load_data_style
+        # TODO: Take first read style for now (make a better system later)
+        src_style = self.load_data_styles[0]
```
mizaki: This could get quite complex I think. An initial idea was to have a window for copying tags?

comictaggerlib/taggerwindow.py
```diff
@@ -1699,7 +1714,8 @@ def copy_tags(self) -> None:
                             failed_list.append(ca.path)
 
                     ca.reset_cache()
-                    ca.load_cache([self.load_data_style, *self.save_data_styles])
+                    # TODO Could result in dupes? Should only be read styles?
+                    ca.load_cache([*self.load_data_styles, *self.save_data_styles])
```
mizaki: Same as above; only read style or all styles?

comictaggerlib/taggerwindow.py
```diff
@@ -1761,10 +1777,13 @@ def metadata_save() -> bool:
         ii = IssueIdentifier(ca, self.config[0], self.current_talker())
 
         # read in metadata, and parse file name if not there
+        # TODO should this follow the same as CLI: filename (-f), read styles (-t), command line (-m)
+        # Duplicated in autotagmatchwindow and renamewindow
+        md = GenericMetadata()
```
mizaki: Will users expect consistency between CLI and GUI? CLI users are more likely to read the CLI options, of course. When `overlyMode` comes into play too... I think `filename` then `styles` and `-m` makes sense unless an option is disabled/empty.

mizaki:
I have a delegate `QListView` version in the works too which would be easier now read styles are kept as a `list`.

Ordering kind of has to come with a list? I altered it so the selection now resets which is more obvious that a change has taken place. Hopefully this helps with the auto-ordering? If not, some idea of how you see it would be helpful. And the order/priority is still outstanding but I'm still of a mind that `1` although it is `last` still "human logically" is the most understandable without knowing the internal mechanics.

lordwelch:
> I'm not sure any sorting is really needed. I don't like that they re-sort whenever I enable one. Also since merging happens as an overlay the "top" item should be the last in the list when it is overlayed as the last item overlayed is what takes priority.

> I went back and forth on the ordering and reasoned that the average user would expect 1 to be the top priority. Naming the column "Priority" (although that takes up extra space) might help? It is "overlay-ed" but I don't think the average user would expect the data in "3" to overwrite the data in "1" etc. Or was it the visual order and not the numbering: "1, 2, 3" (top to bottom) but it should be displayed "3, 2, 1" (top to bottom)?



> Ordering kind of has to come with a list? I altered it so the selection now resets which is more obvious that a change has taken place. Hopefully this helps with the auto-ordering? If not, some idea of how you see it would be helpful. And the order/priority is still outstanding but I'm still of a mind that 1 although it is last still "human logically" is the most understandable without knowing the internal mechanics.

See renamewindow.py line 89 where it isn't reversed

comictaggerlib/taggerwindow.py
lordwelch: It might be better to parse the filename first and overlay the rest of the data but we don't actually provide an option for it in the GUI which we would need to do. But the `-m` option should never be honored in the GUI. Not something to fix on this PR.

comictaggerlib/taggerwindow.py
lordwelch: It should just be a set instead
```python
{*self.load_data_styles, *self.save_data_styles}
```

comictaggerlib/taggerwindow.py
lordwelch: I don't see why. Read all the metadata in, overlay it, and then save it to each metadata type.
You could special case when only a single metadata is selected as a source so that you do one less write operation.

comictaggerlib/taggerwindow.py
lordwelch: Yeah, I didn't update it when I copied it from `adjust_load_style_combo`

comictaggerlib/taggerwindow.py
lordwelch: You removed the code that set it back
```python
        else:
            self.cbLoadDataStyle.currentIndexChanged.disconnect(self.set_load_data_style)
            self.adjust_load_style_combo()
            self.cbLoadDataStyle.currentIndexChanged.connect(self.set_load_data_style)
```

comictaggerlib/renamewindow.py
lordwelch: Since we are adding in multiple read styles yes

comictaggerlib/autotagmatchwindow.py
lordwelch: For auto tagging it might be better to just clear the cache as it should provide at least a small speedup in processing

lordwelch (commented):
Also you still have 20+ commits to rebase out

comictaggerlib/cli.py
mizaki: From my reading of it, the option only takes a single metadata style?
```
parser.add_setting(
        "-c",
        "--copy",
        type=metadata_type_single,
        metavar=f"{{{','.join(metadata_styles).upper()}}}",
        help="Copy the specified source tag block to\ndestination style specified via --type-modify\n(potentially lossy operation).\n\n",
        file=False,
    )
```
So it won't be overlayed?

I was a little unsure on what the "block" meant, if it was only a specific tag e.g. "series" but that doesn't look to be the case. Maybe change to: Copy the specified source tags... ?

comictaggerlib/taggerwindow.py
mizaki: Ah, I see it now. I thought because `currentIndexChanged` wasn't needed to be changed back it could go. Update coming.

comictaggerlib/taggerwindow.py
mizaki: I was thinking of a window so you can choose more independently. I've done the overlayed result so an independent copy tag PR could be a backburner.

mizaki:
I think we're close to just the dreaded tests left to do?

comictaggerlib/taggerwindow.py
```diff
@@ -2151,14 +2169,25 @@ def load_archive(self, comic_archive: ComicArchive) -> None:
...
+            self.exception(f"Failed to load metadata for {self.comic_archive.path}, see log for details\n\n")
+
+        self.update_ui_for_archive()
+
+    def overlay_ca_read_style(self, load_data_styles: list[str], ca: ComicArchive) -> tuple[GenericMetadata, bool]:
+        md = GenericMetadata()
+        success = True
         try:
-            self.metadata = self.comic_archive.read_metadata(self.load_data_style)
+            for style in reversed(load_data_styles):
```
lordwelch: Instead of reversing it here reverse it when setting/getting from the Qt widget.

comictaggerlib/autotagmatchwindow.py
```diff
@@ -229,8 +229,8 @@ def reject(self) -> None:
     def save_match(self) -> None:
         match = self.current_match()
         ca = ComicArchive(self.current_match_set.original_path)
+        md, success = self.parent().overlay_ca_read_style(self.load_data_styles, ca)
```
lordwelch: `success` is never used, either remove the return value from `overlay_ca_read_style` or do something with it here

comictaggerlib/renamewindow.py
```diff
@@ -48,7 +48,9 @@ def __init__(
         with (ui_path / "renamewindow.ui").open(encoding="utf-8") as uifile:
             uic.loadUi(uifile, self)
 
-        self.label.setText(f"Preview (based on {metadata_styles[data_style].name()} tags):")
+        self.label.setText(
+            f"Preview (based on {', '.join([f'{metadata_styles[style].name()}' for style in load_data_styles])} tags):"
```
lordwelch:
```suggestion
            f"Preview (based on {', '.join(metadata_styles[style].name() for style in load_data_styles)} tags):"
```

comictaggerlib/renamewindow.py
```diff
@@ -82,7 +84,7 @@ def config_renamer(self, ca: ComicArchive, md: GenericMetadata | None = None) ->
             new_ext = ca.extension()
 
         if md is None or md.is_empty:
-            md = ca.read_metadata(self.data_style)
+            md, success = self.parent().overlay_ca_read_style(self.load_data_styles, ca)
```
lordwelch: `success` is never used, either remove the return value from `overlay_ca_read_style` or do something with it here

comictaggerlib/taggerwindow.py
```diff
@@ -1203,24 +1205,25 @@ def commit_metadata(self) -> None:
                 self.update_menus()
             self.fileSelectionList.update_current_row()
 
-            self.metadata = self.comic_archive.read_metadata(self.load_data_style)
+            self.metadata, success = self.overlay_ca_read_style(self.load_data_styles, self.comic_archive)
```
lordwelch: `success` is never used, either remove the return value from `overlay_ca_read_style` or do something with it here

comictaggerlib/taggerwindow.py
```diff
@@ -1655,10 +1672,14 @@ def copy_tags(self) -> None:
...
 
-                    if ca.has_metadata(src_style) and ca.is_writable():
-                        md = ca.read_metadata(src_style)
-                    else:
+                    for style in src_styles:
+                        if ca.is_writable() and ca.has_metadata(style):
+                            md.overlay(ca.read_metadata(style))
+                        else:
+                            continue
+                    if md.is_empty:
```
lordwelch: shouldn't this use `overlay_ca_read_style`?

comictaggerlib/taggerwindow.py
```diff
@@ -1727,11 +1748,8 @@ def metadata_save() -> bool:
         ii = IssueIdentifier(ca, self.config[0], self.current_talker())
 
         # read in metadata, and parse file name if not there
-        try:
-            md = ca.read_metadata(self.load_data_style)
-        except Exception as e:
-            md = GenericMetadata()
-            logger.error("Failed to load metadata for %s: %s", ca.path, e)
+        md, success = self.overlay_ca_read_style(self.load_data_styles, ca)
```
lordwelch: `success` is never used, either remove the return value from `overlay_ca_read_style` or do something with it here

comictaggerlib/ctsettings/commandline.py
lordwelch: `-t` should be a short option for `--type-read` and `--type-modify` should should default to `--type-read` (implement it in `validate_commandline_settings`)

lordwelch (changes requested):
I really don't like that it re-arranges when a style gets enabled. Instead of sorting it on click only sort it when the dropdown is opened that way it doesn't jump around.

The rest is a bit of cleanup left that I noticed.

mizaki:
> I really don't like that it re-arranges when a style gets enabled. Instead of sorting it on click only sort it when the dropdown is opened that way it doesn't jump around.
> 
I'm trying to picture it:
```
1. [x] CR
-. [] CoMeT
-. [] CBI
-. [] XML
```
If you click `CBI`:
```
1. [x] CR
-. [] CoMeT
2. [x] CBI
-. [] XML
```
Click `CoMeT`:
```
1. [x] CR
3. [x] CoMeT
2. [x] CBI
-. [] XML
```
Is that how you picture it?

With each click the overlay will be re-done (barring `dirty flag`) so they will need to be ordered correctly behind the scenes.

I think it's probably confusing for the user if they stay unsorted. And it's a "table" so are usually sorted.

And when it comes to moving a style; say move `CBI` up:
```
2. [x] CR
3. [x] CoMeT
1. [x] CBI
-. [] XML
```
Would that be the expected?

I think as long as it's clear to the user "something" has happened. Maybe the moved rows go a colour and fade back? Some animation to see them move?

Taking the number from the label in the `currentData` loop and sorting the order isn't difficult of course, it's more the possible user confusion. To bring back the 90s; WYSIWYG and if the final example is the intent, there are two "orderings" fighting it out. Back to my "priority" (top to bottom) being easier to understand for the average user when it comes to "overlaying".

If they don't move, the hover arrows could be removed and the "Priority Number" could be a spinbox.

comictaggerlib/taggerwindow.py
```diff
@@ -2151,14 +2169,25 @@ def load_archive(self, comic_archive: ComicArchive) -> None:
...
+
+        self.metadata, success = self.overlay_ca_read_style(self.load_data_styles, self.comic_archive)
+        if not success:
+            self.exception(f"Failed to load metadata for {self.comic_archive.path}, see log for details\n\n")
+
+        self.update_ui_for_archive()
+
+    def overlay_ca_read_style(self, load_data_styles: list[str], ca: ComicArchive) -> tuple[GenericMetadata, bool]:
+        md = GenericMetadata()
+        success = True
```
mizaki: The `success` from the `overlay_ca_read_style` is only required because of the exception window created by `load_archive` with an error loading the style. If the `success` return is removed, how would you recommend covering that?

Maybe this isn't the intention for `ComicArchive` but could `self.md` be `None` `dict[str, GenericMetadata | None]` if there is an error with a style?

lordwelch:
If I uncheck a style I don't want it to change. Same if I check a style. Enabling a style is separate from ordering it.

In most if not all cases we don't even need the number column it's redundant with the actual order of the items


If I click CBI I don't want it re-ordered because I'm clicking on CoMet next and it's already in the order I want, I don't want an application that thinks it's smart, unless it actually is, this application isn't smart (I've written a lot of code for it).
```
1. [x] CR
-. [ ] XML
-. [ ] CoMet
-. [ ] CBI
```
Order should stay the same, we have buttons to change it.
```
1. [x] CR
-. [ ] XML
-. [ ] CoMet
2. [x] CBI
```
Now I add CoMet
```
1. [x] CR
-. [ ] XML
2. [x] CoMet
3. [x] CBI
```
Then I close the dialog and I open it back up. Now it shows:
```
1. [x] CR
2. [x] CoMet
3. [x] CBI
-. [ ] XML
```
This may not be a good assumption but it's not an assumption that changes in the middle of me making my choice. It most likely happens after I've done something else and haven't looked at it for a minute, in which case I'll look at all the choices again.

Now with the current behaviour if I have all of them selected:
```
1. [x] CR
2. [x] XML
3. [x] CoMet
4. [x] CBI
```
and I want to deselect 2 and 4 the only way for me to deselect them without the order changing on me as I'm moving my mouse is to click 4 first and then click 2.
```
1. [x] CR
2. [x] XML
3. [x] CoMet
4. [ ] CBI
```
```
1. [x] CR
2. [ ] XML
3. [x] CoMet
4. [ ] CBI
```
If I click 2 and then move down to 4  and click I've now just re-ordered it without deselecting anything. It's a crappy experience because now I have to do each action by itself and wait and see what the stupid application does, because it's not smart.
```
1. [x] CR
2. [x] CoMet
3. [x] CBI
-. [ ] XML
```
```
1. [x] CR
2. [x] CoMet
3. [x] CBI
4. [x] XML
```

comictaggerlib/taggerwindow.py
lordwelch: On line 1751 you have a `success` variable from `overlay_ca_read_style`. Why don't you just stop it there instead of processing a file that you know has an error.

On line 1208 it's loading it back in after it attempted to save if `success` is false that means more than likely that CT broke the archive, or that the archive was broken before, tell the user that an error happened.

mizaki:
> Order should stay the same, we have buttons to change it.
> 
> ```
> 1. [x] CR
> -. [ ] XML
> -. [ ] CoMet
> 2. [x] CBI
> ```
> 
> Now I add CoMet
> 
> ```
> 1. [x] CR
> -. [ ] XML
> 2. [x] CoMet
> 3. [x] CBI
> ```
> 
This was the missing key for me. Numbering them based on their position in the list. 20/20 hindsight etc.

Same with altering the order, they do not move so the numbers will be listed "out of order" but the user put them that way so it's all good (and sorts on dropdown close)?

I think now then that a `spinbox` for the "order" will work better. Just need a bit of logic to alter the other numbers.

In this case it also makes senses to change `itemChanged` to `dropdownClosed` and do the dirty flag check and update the overlay'ed data in the tagger window etc. then?

lordwelch:
> Same with altering the order, they do not move so the numbers will be listed "out of order" but the user put them that way so it's all good (and sorts on dropdown close)?
> 
> I think now then that a `spinbox` for the "order" will work better. Just need a bit of logic to alter the other numbers.

I'm not sure we need to show the number. It already has the order by being able to position the styles. If you really want a number to show the user which item is first, make it based on the row number and update it when an item is moved.
There are only a few cases when being able to move a style to an exact index is going to be useful and there aren't enough separate styles for that.

> In this case it also makes senses to change `itemChanged` to `dropdownClosed` and do the dirty flag check and update the overlay'ed data in the tagger window etc. then?

Yes, you'll have to make your own signal and trigger it when the dropdown is closed

comictaggerlib/taggerwindow.py
mizaki: I was trying to replicate what was already happening with the minimal amount of changes. Maybe I am misunderstanding what is already there:

1751 change. Before it would try and except with an empty `md`, log the error and carry on with an empty GMD. The `overlay_ca_read_style` should be doing the same?

The only time that doesn't happen is the `self.exception` with `load_archive` and why the `success` return as mentioned. Am I missing something that was happening before or should the `logger` exceptions not be "quiet"? Or:

```python
    def overlay_ca_read_style(self, load_data_styles: list[str], ca: ComicArchive) -> tuple[GenericMetadata, dict | None]:
        md = GenericMetadata()
        error = None
        try:
            for style in load_data_styles:
                metadata = ca.read_metadata(style)
                md.overlay(metadata)
        except Exception as e:
            error = e

        return md, error
```
And then the thing calling `overlay_ca_read_style` logs/displays the message?

```python
md, error = self.overlay_ca_read_style(self.load_data_styles, self.comic_archive)
        if error:
            logger.error("Failed to load metadata for %s: %s", self.ca.path, error)
            # additional if required
            self.exception(f"Failed to load metadata for {self.ca.path}: {error}\n\n")
```
I was trying to leave small footsteps.

lordwelch: > I was trying to replicate what was already happening with the minimal amount of changes

There's your problem. Don't try to keep everything exactly as it was. You created a new function make use of it the best that you can. If it needs to change then change the created function. If the calling function needs to change then change the calling function. With this project don't assume that the previous code was done the "correct way" it's just the code that happened to be there, maybe it's correct maybe it's not.

In a PR if you create a new function to consolidate code and you give it a "success" return value I would expect it to be used not, arbitrarily ignored.

mizaki:
Using item delegate:
image
Still a few small tweaks could be done but overall it's mostly there.

It requires a little merry-go-round with signals but otherwise it gets to a custom rolled view/model to just re-implement what the combox already does.

I believe for you the modify style dropdown checkbox are the normal size but for me they are massive. Not sure on windows so a unifying pass is probably needed.

The movement arrows are only shown for `checked` items but it may be as well to show them for all.

comictaggerlib/autotagmatchwindow.py
```diff
@@ -229,8 +229,17 @@ def reject(self) -> None:
...
+        md, error = self.parent().overlay_ca_read_style(self.load_data_styles, ca)
+        if error is not None:
+            logger.error("Failed to load metadata for %s: %s", ca.path, error)
+            QtWidgets.QApplication.restoreOverrideCursor()
+            QtWidgets.QMessageBox.critical(
+                self,
+                "Read Failed!",
+                f"One or more of the read styles failed to load for {ca.path}, check log for details",
+            )
+            return
```
mizaki, line 241: Now quit the save on an error

comictaggerlib/renamewindow.py
```diff
@@ -82,7 +84,11 @@ def config_renamer(self, ca: ComicArchive, md: GenericMetadata | None = None) ->
             new_ext = ca.extension()
 
         if md is None or md.is_empty:
-            md = ca.read_metadata(self.data_style)
+            md, error = self.parent().overlay_ca_read_style(self.load_data_styles, ca)
+            # How much should we care about this error? Notify user or not?
+            if error is not None:
+                logger.error("Failed to load metadata for %s: %s", ca.path, error)
+
```
mizaki: Should the user be informed in a window or is it not all that relevant here?

comictaggerlib/taggerwindow.py
```diff
@@ -1655,10 +1681,11 @@ def copy_tags(self) -> None:
...
                     ca_saved = False
-
-                    if ca.has_metadata(src_style) and ca.is_writable():
-                        md = ca.read_metadata(src_style)
-                    else:
+                    md, error = self.overlay_ca_read_style(src_styles, ca)
+                    if error is not None:
+                        failed_list.append(ca.path)
+                        continue
+                    if md.is_empty:
```
mizaki, line 1690: The `failed_list` is used for writing styles but as a failed read will likely also be a failed write, as per the same with the write error, bail on this archive right away.

comictaggerlib/ui/customwidgets.py
```diff
@@ -112,3 +122,256 @@ def toggleItem(self, index: int) -> None:
...
+        painter.drawText(rectangle, Qt.AlignVCenter, label)
+
+        # Draw buttons
+        if checked and (option.state & QtWidgets.QStyle.State_Selected):
+            # TODO Fill button rects or similar for better separation from text below?
+            up_rect = self._button_up_rect(option.rect)
+            down_rect = self._button_down_rect(option.rect)
+
+            painter.drawImage(up_rect, QtGui.QImage(str(graphics_path / "up.png")))
+            painter.drawImage(down_rect, QtGui.QImage(str(graphics_path / "down.png")))
```
mizaki: At the moment the icon arrows can get a bit lost with any text underneath. Adding a little extra length to the `sizeHint` or `setMinimumContentsLength` might be a way around it.

comictaggerlib/ui/customwidgets.py
```diff
@@ -112,3 +122,256 @@ def toggleItem(self, index: int) -> None:
...
+            # We record that the combobox list has been shown
+            if event.type() == QEvent.Show:
+                self.justShown = True
+            # We record that the combobox list has hidden,
+            # this will happen if the user does not make a selection
+            # but clicks outside of the combobox list or presses escape
+            if event.type() == QEvent.Hide:
+                self._updateText()
+                self.justShown = False
+                self.dropdownClosed.emit(self.currentData())
```
mizaki, line 319: I thought this made sense. Doing the same for modify would as well. Could also change modify to dropdown close too instead of item check change. Although I do like seeing the fields open and close when clicking the modify styles on and off.

mizaki:
The `CheckableOrderComboBox` I subbed from `QComboBox` for initial ease. I'm not sure if much would be gained from sub-classing `CheckableComboBox` instead?

comictaggerlib/renamewindow.py
lordwelch: The user should have already been informed but we should still let them know here

comictaggerlib/ui/customwidgets.py
lordwelch: You should just be able to add up the width of all the widgets and give it in the sizeHint

lordwelch:
Nope their also bigger for me
Screenshot 2024-05-06 at 6 39 38 PM

But it also doesn't match the style of the other combobox
Screenshot 2024-05-06 at 6 56 25 PM

mizaki:
> Nope their also bigger for me

Seems to be [known](https://stackoverflow.com/a/20118475). Wonder about PyQt6...? Need to switch to 6 or work around the combobox placeholder text issue too.

Use the same delegate for both modify and read and hide the arrows might be the best answer?

mizaki:
PyQt6 does do the same drawing the checkbox differently so I'll unify `CheckableComboBox` and `CheckableOrderComboBox` to look the same.

lordwelch:
So I found out why it's different.
The combobox uses a menu for displaying the items by default https://github.com/qt/qtbase/blob/9a61bc5dfc3c68200dbf48fd79771a856fae26d5/src/widgets/widgets/qcombobox_p.h#L237
For macos they use a menu check https://github.com/qt/qtbase/blob/9a61bc5dfc3c68200dbf48fd79771a856fae26d5/src/plugins/styles/mac/qmacstyle_mac.mm#L4164C1-L4185C14

Here's a patch. I also removed the sizeHint and it went down to a normal size.
[checkbox-fix.patch](https://github.com/comictagger/comictagger/files/15256203/checkbox-fix.patch)

mizaki:
> So I found out why it's different. The combobox uses a menu for displaying the items by default https://github.com/qt/qtbase/blob/9a61bc5dfc3c68200dbf48fd79771a856fae26d5/src/widgets/widgets/qcombobox_p.h#L237 For macos they use a menu check https://github.com/qt/qtbase/blob/9a61bc5dfc3c68200dbf48fd79771a856fae26d5/src/plugins/styles/mac/qmacstyle_mac.mm#L4164C1-L4185C14
> 
> Here's a patch. I also removed the sizeHint and it went down to a normal size. [checkbox-fix.patch](https://github.com/comictagger/comictagger/files/15256203/checkbox-fix.patch)

The `CheckableComboBox` and `CheckableOrderComboBox` both look the same for you now?

My issue was `CheckableComboBox` making the checkbox big:
image
(And going by the cut off text the `sizeHint` is using the small/correct size box for the calc.)

Using `CheckableOrderComboBox` gives me this by default (which is the "correct" size checkbox):
image

comictaggerlib/taggerwindow.py
```diff
@@ -211,18 +211,20 @@ def __init__(
...
             if style not in metadata_styles:
                 config[0].internal__save_data_style.remove(style)
-        if config[0].internal__load_data_style not in metadata_styles:
-            config[0].internal__load_data_style = list(metadata_styles.keys())[0]
+        for style in config[0].internal__load_data_style:
+            if style not in metadata_styles:
+                config[0].internal__load_data_style.remove(style)
         self.save_data_styles: list[str] = config[0].internal__save_data_style
-        self.load_data_style: str = config[0].internal__load_data_style
+        self.load_data_styles: list[str] = list(reversed(config[0].internal__load_data_style))
```
lordwelch: This shouldn't be reversed here. Why is it being saved in the wrong order

comictaggerlib/taggerwindow.py
```diff
@@ -2065,6 +2097,7 @@ def closeEvent(self, event: QtGui.QCloseEvent) -> None:
                 self.config[0].internal__sort_column,
                 self.config[0].internal__sort_direction,
             ) = self.fileSelectionList.get_sorting()
+            self.config[0].internal__load_data_style = self.cbLoadDataStyle.currentData()
```
lordwelch: Why is this added here? why isn't this set in the equivalent place that the save data style is set

comictaggerlib/taggerwindow.py
```diff
@@ -1201,26 +1203,35 @@ def commit_metadata(self) -> None:
...
 
-    def set_load_data_style(self, s: str) -> None:
+    def set_load_data_style(self, load_data_styles: list[str]) -> None:
         if self.dirty_flag_verification(
-            "Change Tag Read Style", "If you change read tag style now, data in the form will be lost.  Are you sure?"
+            "Change Tag Read Style",
+            "If you change read tag style(s) now, data in the form will be lost.  Are you sure?",
         ):
-            self.load_data_style = self.cbLoadDataStyle.itemData(s)
-            self.config[0].internal__load_data_style = self.load_data_style
```
lordwelch, line 1216: Why was this removed?

lordwelch (changes requested):
I think the best option would be to re-create QComboMenuDelegate in python and then add the buttons onto the end and extend the sizehint width to include the buttons.

But I don't really care I'll merge how it looks as is.
There are 4 comments that I want taken care of first before I merge.

comictaggerlib/taggerwindow.py
mizaki: Reverted.

It was because the `self.config[0].internal__load_data_style` is only used when saving and it had to be unreversed and so was set on `closeEvent` it didn't need to be set. Thinking on it, that may have caused issues with the settings window saving.

comictaggerlib/taggerwindow.py
mizaki: Reverted.

This was because of the whole "reversed" thing.

comictaggerlib/taggerwindow.py
mizaki: Reverted.

I thought it made sense to save in the user facing order but it doesn't really.

lordwelch:
The only last thing is the calls to `reversed` it should be consistent you need to have the reversed calls either both in customwidgets.py or both in taggerwindow.py. Putting them both in taggerwindow.py and putting a comment in `set_load_data_style` and `adjust_load_style_combo` explaining why it's reversed should be good. I mainly don't want anyone in 6 months or longer looking at this and getting confused about why the list gets reversed on them

mizaki:
> The only last thing is the calls to `reversed` it should be consistent you need to have the reversed calls either both in customwidgets.py or both in taggerwindow.py. Putting them both in taggerwindow.py and putting a comment in `set_load_data_style` and `adjust_load_style_combo` explaining why it's reversed should be good. I mainly don't want anyone in 6 months or longer looking at this and getting confused about why the list gets reversed on them

I used your comments and reverse only in taggerwindow. I didn't use the `eventClose` for `self.config[0].internal__load_data_style = self.load_data_styles` and left it in `set_load_data_style` because I think it'll cause problems with saving of settings in the settings window (using the old value).

lordwelch (approved):

PR #642, opened 2024-05-10
Use custom delegate to unify combobox item style

mizaki:
On Linux (tested) changes:
image
Which are oversized checkboxes to:
image

### DIFF
--- a/comictaggerlib/ui/customwidgets.py
+++ b/comictaggerlib/ui/customwidgets.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 from enum import auto
 from sys import platform
-import platform
 from typing import Any
 
 from PyQt5 import QtGui, QtWidgets
--- a/comictaggerlib/ui/customwidgets.py
+++ b/comictaggerlib/ui/customwidgets.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 from enum import auto
 from sys import platform
+import platform
 from typing import Any
 
 from PyQt5 import QtGui, QtWidgets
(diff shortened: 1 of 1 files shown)


mizaki:
A combined delegate for read and modify could be worked out (signals etc.) but for ease right now.

lordwelch:
Now it's even worse on MacOS....
Screenshot 2024-06-21 at 7 14 52 PM

mizaki:
> Now it's even worse on MacOS.... Screenshot 2024-06-21 at 7 14 52 PM

So it's an extra tick but is the tick also a bit larger too?

I was wondering if PyQt6 is going to make any difference to this at all but I don't think it will.

IIRC the problem was only on Linux (I'll check), so I think it's possible to do this change only for Linux.

mizaki:
Try this. It looks like the "tick" was being drawn twice (didn't notice because it was perfectly aligned) so I've removed the draw from the code here.

lordwelch:
Now I don't get any visible check on MacOS and It crashed when I clicked on the 'Write Tags' checkbox too much
```plain
2024-07-27 14:14:05.686 Python[23532:673543] WARNING: Secure coding is not enabled for restorable state! Enable secure coding by implementing NSApplicationDelegate.applicationSupportsSecureRestorableState: and returning YES.
2024-07-27T14:14:23 | comictagger | CRITICAL | Uncaught exception: AttributeError: 'NoneType' object has no attribute 'checkState'
  File "/Users/timmy/build/source/comictagger/comictaggerlib/ui/customwidgets.py", line 144, in eventFilter
    self.toggleItem(index.row())
  File "/Users/timmy/build/source/comictagger/comictaggerlib/ui/customwidgets.py", line 205, in toggleItem
    if self.model().item(index).checkState() == Qt.Checked:
AttributeError: 'NoneType' object has no attribute 'checkState'
```

mizaki:
Try this. If not then skipping it on MacOS might be the option. I did wonder about the purpose of `if platform != "darwin":` and so left it in. I thought it might only draw the tick but the whole check box was missing?

Not too sure on the `None` as it shouldn't be related but let's see if it goes away once the box is drawn again.

lordwelch:
By default on MacOS it only has a tick. There is no checkbox

mizaki:
> By default on MacOS it only has a tick. There is no checkbox

In the double tick screenshot you gave above; is the box there not usually there?

Does this just draw a box and the tick or just the box? Any harm in having the boxes?

If there's still a problem: Because I can't try different solution, if I put in some different options commented out so you can try them, is that okay?

lordwelch:
> If there's still a problem: Because I can't try different solution, if I put in some different options commented out so you can try them, is that okay?

Sure

lordwelch:
It looks like you can reset to a44e0373110b04dad5e2997228463e56a539c605 and then you just need to remove the extra import platform

mizaki:
> It looks like you can reset to [a44e037](https://github.com/comictagger/comictagger/commit/a44e0373110b04dad5e2997228463e56a539c605) and then you just need to remove the extra import platform

I've reset and removed the double import. Is that original one now working (with the PyQt5 5.15.11 update)?

If not, I'll push the other options.

lordwelch:
It works on MacOS now

PR #646, opened 2024-05-23
Add read and source merge list options for overlay

mizaki:
Not sure on the labeling:
image
`Non-lists` could use a better term?


lordwelch:
I'm going with #616 instead as I fix other issues at the same time

PR #653, opened 2024-06-01
Add remove HTML tables back

mizaki:
It seems the option got lost along the way.


lordwelch (approved):
It got removed in #430

PR #693, opened 2024-10-03
Preserve hidden metadata values when reading from GUI form

mizaki:
Currently when `write_tags` is used, `form_to_metadata` is called within it which means any "hidden" values such as `data-origin`, `identifier`, `price`, `rights`, `issue_id`, `series_id`, `is_version_of` and `last_mark` are erased.

Some of these values are/may be of use to a tag plugin. (Saving the `issue_is` and `series_id` in ACBF for example.)

I'm not sure if doing it in the `form_to_metadata` will cause problems anywhere else? The other option I considered was to pull the values out in the `write_tags` method and then place them back after `form_to_metadata`.

### DIFF
--- a/comictaggerlib/taggerwindow.py
+++ b/comictaggerlib/taggerwindow.py
@@ -940,6 +940,16 @@ class TaggerWindow(QtWidgets.QMainWindow):
             md.add_credit(name, role, bool(primary_flag))
 
         md.pages = self.page_list_editor.get_page_list()
+
+        # Preserve hidden md values
+        md.data_origin = self.metadata.data_origin
+        md.issue_id = self.metadata.issue_id
+        md.series_id = self.metadata.series_id
+
+        md.price = self.metadata.price
+        md.identifier = self.metadata.identifier
+        md.rights = self.metadata.rights
+
         self.metadata = md
 
     def use_filename(self) -> None:


comictaggerlib/taggerwindow.py
```diff
@@ -940,6 +940,9 @@ def form_to_metadata(self) -> None:
             md.add_credit(name, role, bool(primary_flag))
 
         md.pages = self.page_list_editor.get_page_list()
+
+        # Preserve hidden md values
+        md.overlay(self.metadata, mode=comicapi.merge.Mode.ADD_MISSING, merge_lists=False)
```
lordwelch: This doesn't account for when someone wants to remove something, how would I remove the title from a comic that already has one set

lordwelch (changes requested):

mizaki:
I knew there was a reason why I rejected that way initially...

I've left it in the `form_to_metadata` method because `apply_cbl_transform` and `auto_imprint` could also wipe those values. Other instances of `form_to_metadata` shouldn't make a difference.

As the tag plugins now (I believe all?) receive the initial read data, it might be better to add `price` and `identifier` to the GUI and remove `is_version_of`, `rights` and `last_mark` from `GenericMetadata` (as a separate PR)?

comictaggerlib/taggerwindow.py
```diff
@@ -940,8 +940,27 @@ def form_to_metadata(self) -> None:
...
+
+        # Preserve hidden md values
+        origin = self.metadata.data_origin
+        issue_id = self.metadata.issue_id
+        series_id = self.metadata.series_id
+
+        price = self.metadata.price
+        identifier = self.metadata.identifier
+        rights = self.metadata.rights
+
```
lordwelch:
```suggestion

        # Preserve hidden md values
        md.origin = self.metadata.data_origin
        md.issue_id = self.metadata.issue_id
        md.series_id = self.metadata.series_id

        md.price = self.metadata.price
        md.identifier = self.metadata.identifier
        md.rights = self.metadata.rights

```

comictaggerlib/taggerwindow.py
```diff
@@ -940,8 +940,27 @@ def form_to_metadata(self) -> None:
...
 
+        # Reapply hidden values
+        self.metadata.data_origin = origin
+        self.metadata.issue_id = issue_id
+        self.metadata.series_id = series_id
+
+        self.metadata.price = price
+        self.metadata.identifier = identifier
+        self.metadata.rights = rights
+
```
lordwelch:
```suggestion
```

lordwelch (changes requested):

lordwelch (approved):

PR #718, opened 2025-01-10
Use language ISO to from credit window

### LINKED ISSUE #711: Credit language
Currently the full name is taken `self.cbLanguage.currentText()` but the `genericmetadata` is supposed to be ISO. If 
```python
    def get_credit(self) -> Credit:
        return Credit(
            self.leName.text(), self.cbRole.currentText(), self.cbPrimary.isChecked(), self.cbLanguage.currentText()
        )
```
is changed to `self.leName.text(), self.cbRole.currentText(), self.cbPrimary.isChecked(), self.cbLanguage.currentData()` the `TaggerWindow` text then shows only the two letter ISO code.

For a PR, either change the `CreditEditorWindow` `get_credit` to return `Credit` and the full language:
```python
    def get_credit(self) -> tuple[Credit, str]:
        return Credit(
            self.leName.text(), self.cbRole.currentText(), self.cbPrimary.isChecked(), self.cbLanguage.currentData()
        ), self.cbLanguage.currentText()
```
and then add in `TaggerWindow`:
```python
        if editor.result():
            new, full_lang = editor.get_credit()
self.twCredits.item(row, self.md_attributes["credits.language"]).setText(full_lang)
self.twCredits.item(row, self.md_attributes["credits.language"]).setData(new.language)
```

OR do the same as setting when reading from metadata:
```python
self.twCredits.item(row, self.md_attributes["credits.language"]).setText(utils.get_language_from_iso(new.language) or new.language)
self.twCredits.item(row, self.md_attributes["credits.language"]).setData(new.language)
```

`form_to_metadata` is then changed to `lang = self.twCredits.item(row, self.md_attributes["credits.language"]).data()` of course.

Which is preferred, he latter `utils` way?

mizaki:
Pass credit language ISO using the widget.data to respect the metadata credit requiring an ISO string. If the string fails to match an ISO, use the raw text.

Fixes: #711

### DIFF
--- a/comictaggerlib/crediteditorwindow.py
+++ b/comictaggerlib/crediteditorwindow.py
@@ -75,8 +75,10 @@ class CreditEditorWindow(QtWidgets.QDialog):
                 self.cbRole.setCurrentIndex(i)
 
         if credit.language != "":
-            i = self.cbLanguage.findData(credit.language, QtCore.Qt.ItemDataRole.UserRole) or self.cbLanguage.findText(
-                credit.language
+            i = (
+                self.cbLanguage.findData(credit.language, QtCore.Qt.ItemDataRole.UserRole)
+                if self.cbLanguage.findData(credit.language, QtCore.Qt.ItemDataRole.UserRole) > -1
+                else self.cbLanguage.findText(credit.language)
             )
             if i == -1:
                 self.cbLanguage.setEditText(credit.language)
--- a/comictaggerlib/crediteditorwindow.py
+++ b/comictaggerlib/crediteditorwindow.py
@@ -20,7 +20,7 @@ import logging
 import operator
 
 import natsort
-from PyQt5 import QtWidgets, uic
+from PyQt5 import QtCore, QtWidgets, uic
 
 from comicapi import utils
 from comicapi.genericmetadata import Credit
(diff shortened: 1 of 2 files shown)


comictaggerlib/crediteditorwindow.py
```diff
@@ -75,7 +75,9 @@ def __init__(self, parent: QtWidgets.QWidget, mode: int, credit: Credit) -> None
                 self.cbRole.setCurrentIndex(i)
 
         if credit.language != "":
-            i = self.cbLanguage.findText(credit.language)
+            i = self.cbLanguage.findData(credit.language, QtCore.Qt.ItemDataRole.UserRole) or self.cbLanguage.findText(
+                credit.language
+            )
```
lordwelch: If findData doesn't find something it returns -1 which is truthy `bool(-1) is True` so this won't work you have to separate it out into it's own if statement

lordwelch (changes requested):

comictaggerlib/crediteditorwindow.py
mizaki: I always forget -1 isn't false...

lordwelch:
I didn't like your fix, undo your last commit and I'll merge this

mizaki:
> I didn't like your fix, undo your last commit and I'll merge this

Done

lordwelch (approved):

PR #730, opened 2025-02-11
Use source hashes for cover matching

mizaki:
Is this the kind of thing you had in mind?

This turned out quite simple.

### DIFF
--- a/comictaggerlib/cli.py
+++ b/comictaggerlib/cli.py
@@ -639,7 +639,7 @@ class CLI:
                             month=ct_md.month,
                             year=ct_md.year,
                             publisher=None,
-                            image_url=ct_md._cover_image or "",
+                            image_url=str(ct_md._cover_image) or "",
                             alt_image_urls=[],
                             description=ct_md.description or "",
                         )
--- a/comictaggerlib/issueidentifier.py
+++ b/comictaggerlib/issueidentifier.py
@@ -307,7 +307,7 @@ class IssueIdentifier:
     def _get_issue_cover_match_score(
         self,
         primary_img_url: str | ImageHash,
-        alt_urls: list[str],
+        alt_urls: list[str | ImageHash],
         local_hashes: list[tuple[str, int]],
         use_alt_urls: bool = False,
     ) -> Score:
@@ -316,17 +316,23 @@ class IssueIdentifier:
 
         # If there is no URL return 100
         if not primary_img_url:
-            return Score(score=100, url="", remote_hash=0)
+            return Score(score=100, url="", remote_hash=0, local_hash=0, local_hash_name="0")
 
         self._user_canceled()
 
+        remote_hashes = []
+        # If the cover is ImageHash and the alternate covers are URLs, the alts will not be hashed/checked currently
         if isinstance(primary_img_url, ImageHash):
-            remote_hashes = [("0", primary_img_url.Hash)]
+            # ImageHash doesn't have a url so we just give it an empty string
+            remote_hashes.append(("", primary_img_url.Hash))
+            if use_alt_urls and alt_urls:
+                remote_hashes.extend(("", alt_hash.Hash) for alt_hash in alt_urls if isinstance(alt_hash, ImageHash))
         else:
             urls = [primary_img_url]
             if use_alt_urls:
-                urls.extend(alt_urls)
-                self.log_msg(f"[{len(alt_urls)} alt. covers]")
+                only_urls = [url for url in alt_urls if isinstance(url, str)]
+                urls.extend(only_urls)
+                self.log_msg(f"[{len(only_urls)} alt. covers]")
 
             remote_hashes = self._get_remote_hashes(urls)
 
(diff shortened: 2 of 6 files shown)


mizaki:
Metron talker patch:
```
Subject: [PATCH] Use image hash
---
Index: metron_talker/metron.py
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/metron_talker/metron.py b/metron_talker/metron.py
--- a/metron_talker/metron.py	(revision 4e0d5c134d65ac4aa908a1cbd45a71cc6846d2c7)
+++ b/metron_talker/metron.py	(date 1739233743602)
@@ -31,7 +31,7 @@
 
 import settngs
 from comicapi import utils
-from comicapi.genericmetadata import ComicSeries, GenericMetadata, MetadataOrigin
+from comicapi.genericmetadata import ComicSeries, GenericMetadata, MetadataOrigin, ImageHash
 from comicapi.issuestring import IssueString
 from comicapi.utils import LocationParseError, parse_url
 from comictalker import talker_utils
@@ -1022,7 +1022,8 @@
             elif series["year_began"]:
                 md.year = utils.xlate_int(series["year_began"])
 
-        md._cover_image = issue["image"]
+        # md._cover_image = issue["image"]
+        md._cover_image = ImageHash(Hash=int(issue["cover_hash"], 16), Kind="phash")
 
         if issue.get("alt_number"):
             md.alternate_number = issue["alt_number"]
```

comictaggerlib/issueidentifier.py
```diff
@@ -620,6 +623,16 @@ def _cover_matching(
         extra_images: list[tuple[str, Image.Image]],
         issues: list[tuple[ComicSeries, GenericMetadata]],
     ) -> list[IssueResult]:
+        # Set hashing kind, will presume all hashes are of the same kind
+        for series, issue in issues:
+            if isinstance(issue._cover_image, ImageHash):
+                if issue._cover_image.Kind == "phash":
+                    self.image_hasher = 3
+                    break
+                elif issue._cover_image == "ahash":
```
lordwelch: Missing `.Kind`?

comictaggerlib/issueidentifier.py
```diff
@@ -536,7 +539,7 @@ def _match_covers(
                 month=issue.month,
                 year=issue.year,
                 publisher=None,
-                image_url=image_url,
```
lordwelch: Lets set this to the empty string so we don't get errors elsewhere.

lordwelch (commented):
Yes, this looks good

comictaggerlib/issueidentifier.py
```diff
@@ -320,12 +320,15 @@ def _get_issue_cover_match_score(
 
         self._user_canceled()
 
-        urls = [primary_img_url]
-        if use_alt_urls:
-            urls.extend(alt_urls)
-            self.log_msg(f"[{len(alt_urls)} alt. covers]")
+        if isinstance(primary_img_url, ImageHash):
+            remote_hashes = [("0", primary_img_url.Hash)]
```
lordwelch: We could also make the alt hashes optionally ImageHashes as well, but this could also be done later

mizaki:
I don't think it's going to be much more effort to put a `cover_hashes` and `alt_covers_hashes` into `GenericMetadata` and then it's all set up and ready to go into the future? (And also, the issue cover images are broken in the issue window which there isn't a real way around if the url has been replaced.)

Only real question is representation of the GM field: `tuple[ImageHash]` or `{phash: <hash>, ahash: <hash>}`?

lordwelch:
> I don't think it's going to be much more effort to put a cover_hashes and alt_covers_hashes into GenericMetadata and then it's all set up and ready to go into the future?

Lets keep it as is for now. We can look at other ways of doing it later

> (And also, the issue cover images are broken in the issue window which there isn't a real way around if the url has been replaced.)

https://github.com/comictagger/comictagger/pull/730#discussion_r1950260237 should take care of any unexpected errors, they should already handle an empty url (or the metron talker shouldn't work there at all).

mizaki:
> > I don't think it's going to be much more effort to put a cover_hashes and alt_covers_hashes into GenericMetadata and then it's all set up and ready to go into the future?
> 
> Lets keep it as is for now. We can look at other ways of doing it later
> 

I've added `URL` to `ImageHash` which seems like the simplest options atm and solves the problems.

If you're happy, the hard part next is making the tests :)

lordwelch:
It's simpler to just explicitly ignore the ImageHash outside of the IssueIdentifier

lordwelch (approved):

comictaggerlib/cli.py
```diff
@@ -639,7 +639,7 @@ def save(self, ca: ComicArchive, match_results: OnlineMatchResults) -> tuple[Res
                             month=ct_md.month,
                             year=ct_md.year,
                             publisher=None,
-                            image_url=ct_md._cover_image or "",
+                            image_url=str(ct_md._cover_image) or "",
```
mizaki, line 642: Is this okay as is or should it be `ct_md._cover_image if isinstance(issue._cover_image, str) else ""` as well?

mizaki:
I thought adding the URL to `ImageHash` covered all the bases and at a later date it could be used as a fallback if wanted. But I can see that not being wanted anyway.

Is there anything you want me to do with this?

PR #734, opened 2025-02-28
Change failed image loading from logger exception to warning

### LINKED ISSUE #732: CoverWidget exception with empty data
```
  File "/home/workspace/comictagger/comictaggerlib/ui/qtutils.py", line 133, in get_qimage_from_data
    Image.open(io.BytesIO(image_data)).save(buffer, format="ppm")
  File "/home/workspace/comictagger/venv/lib64/python3.10/site-packages/PIL/Image.py", line 3298, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7f97a23f8450>
2025-02-27T15:29:12 | comictaggerlib.ui.qtutils | ERROR | Failed to load the image
  File "/home/workspace/comictagger/comictaggerlib/ui/qtutils.py", line 133, in get_qimage_from_data
    Image.open(io.BytesIO(image_data)).save(buffer, format="ppm")
  File "/home/workspace/comictagger/venv/lib64/python3.10/site-packages/PIL/Image.py", line 3298, in open
    raise UnidentifiedImageError(msg)
PIL.UnidentifiedImageError: cannot identify image file <_io.BytesIO object at 0x7f97a23fa520>
```

The options I see are:
1. Add a `URL` to `ImageHash` to be used for the issue window
2. If the `_cover_image` is an `ImageHash`, do nothing (default blank)

mizaki:
Could also change to `debug` or `pass` it?

Fixes #732

### DIFF
--- a/comictaggerlib/ui/qtutils.py
+++ b/comictaggerlib/ui/qtutils.py
@@ -133,7 +133,7 @@ if qt_available:
                     Image.open(io.BytesIO(image_data)).save(buffer, format="ppm")
                     success = img.loadFromData(buffer.getvalue())
             except Exception:
-                logger.exception("Failed to load the image")
+                logger.warning("Failed to load the image")
         # if still nothing, go with default image
         if not success:
             img.load(":/graphics/nocover.png")
--- a/comictaggerlib/ui/qtutils.py
+++ b/comictaggerlib/ui/qtutils.py
@@ -133,7 +133,7 @@ if qt_available:
                     Image.open(io.BytesIO(image_data)).save(buffer, format="ppm")
                     success = img.loadFromData(buffer.getvalue())
             except Exception:
-                logger.warning("Failed to load the image")
+                logger.exception("Failed to load the image. Size: %s", len(image_data))
         # if still nothing, go with default image
         if not success:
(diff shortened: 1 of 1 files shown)


lordwelch:
I don't understand why you want this. This just changes what gets logged no functionality changes and now we don't get a stack trace

mizaki:
The problem is it will try to load empty bytes.

The proper fix then is probably to test for 0 bytes first?

mizaki:
It doesn't crash CT out but you get the log as in #732 which to my mind, isn't a real exception.

Should 0 bytes not get that far? It seems like a reasonable check to do though?

lordwelch:
Here I think it is an exception, callers should check that they are sending non empty bytes. You can put a check in to log it with a proper message but it should still be logged as an error

mizaki:
Added the size to the logger message. Hopefully that should be enough to not go chasing ghosts.

lordwelch:
Instead put a check at the beginning to return early and log that an empty image was provided

lordwelch (approved):

mizaki:
Ready for merge?

PR #735, opened 2025-02-28
Add URL to ImageHash and use in issue window

mizaki:
This allows to use of `ImageHash` and for the issue window to display covers.

`alt_images = [str(url) for url in issue._alternate_images]` I didn't roll into `[str(issue._cover_image) or "", *alt_images]` for ease of reading.

### DIFF
--- a/comictaggerlib/issueidentifier.py
+++ b/comictaggerlib/issueidentifier.py
@@ -314,40 +314,34 @@ class IssueIdentifier:
         # local_hashes is a list of pre-calculated hashes.
         # use_alt_urls - indicates to use alternate covers
 
-        # If there is no ImageHash or no URL and Hash, return 100 for a bad match
-        if primary_img_url is None or (primary_img_url.Hash == 0 and not primary_img_url.URL and not use_alt_urls):
+        # If there is no ImageHash or no URL and Kind, return 100 for a bad match
+        if primary_img_url is None or (not primary_img_url.Kind and not primary_img_url.URL and not use_alt_urls):
             return Score(score=100, url="", remote_hash=0, local_hash=0, local_hash_name="0")
 
         self._user_canceled()
 
         remote_hashes = []
 
-        try:
-            if primary_img_url.Hash != 0 and primary_img_url.Kind:
-                remote_hashes.append((primary_img_url.URL or "", primary_img_url.Hash))
-                self.log_msg(
-                    f"Using source talker hash for cover matching. Hash: {primary_img_url.Hash}, Kind: {primary_img_url.Kind}"
-                )
-            elif primary_img_url.URL:
-                remote_hashes = self._get_remote_hashes([primary_img_url.URL])
-                self.log_msg("Using URL download image for cover matching.")
-        except Exception as e:
-            logger.warning("Failed to process primary ImageHash: %s", e)
+        if primary_img_url.Kind:
+            remote_hashes.append((primary_img_url.URL, primary_img_url.Hash))
+            self.log_msg(
+                f"Using provided hash for cover matching. Hash: {primary_img_url.Hash}, Kind: {primary_img_url.Kind}"
+            )
+        elif primary_img_url.URL:
+            remote_hashes = self._get_remote_hashes([primary_img_url.URL])
+            self.log_msg(f"Downloading image for cover matching: {primary_img_url.URL}")
 
         if use_alt_urls and alt_urls:
-            try:
-                only_urls = []
-                for alt_url in alt_urls:
-                    if alt_url.Hash != 0 and alt_url.Kind:
-                        remote_hashes.append((alt_url.URL or "", alt_url.Hash))
-                    elif alt_url.URL:
-                        only_urls.append(alt_url.URL)
-                if only_urls:
-                    remote_hashes.extend(self._get_remote_hashes(only_urls))
+            only_urls = []
+            for alt_url in alt_urls:
+                if alt_url.Kind:
+                    remote_hashes.append((alt_url.URL, alt_url.Hash))
+                elif alt_url.URL:
+                    only_urls.append(alt_url.URL)
+            if only_urls:
+                remote_hashes.extend(self._get_remote_hashes(only_urls))
 
-                self.log_msg(f"[{len(remote_hashes) - 1} alt. covers]")
-            except Exception as e:
-                logger.warning("Failed to process alternative ImageHash(s): %s", e)
+            self.log_msg(f"[{len(remote_hashes) - 1} alt. covers]")
 
         score_list = []
         done = False
@@ -540,7 +534,7 @@ class IssueIdentifier:
             try:
                 # We only include urls in the IssueResult so we don't have to deal with it down the line
                 # TODO: display the hash to the user so they know a direct hash was used instead of downloading an image
-                alt_urls: list[str] = [img.URL or "" for img in issue._alternate_images]
+                alt_urls: list[str] = [img.URL for img in issue._alternate_images]
 
                 score_item = self._get_issue_cover_match_score(
                     issue._cover_image, issue._alternate_images, hashes, use_alt_urls=use_alternates
--- a/comictaggerlib/issueidentifier.py
+++ b/comictaggerlib/issueidentifier.py
@@ -306,35 +306,48 @@ class IssueIdentifier:
 
     def _get_issue_cover_match_score(
         self,
-        primary_img_url: str | ImageHash,
-        alt_urls: list[str | ImageHash],
+        primary_img_url: ImageHash | None,
+        alt_urls: list[ImageHash],
         local_hashes: list[tuple[str, int]],
         use_alt_urls: bool = False,
     ) -> Score:
         # local_hashes is a list of pre-calculated hashes.
-        # use_alt_urls - indicates to use alternate covers from CV
+        # use_alt_urls - indicates to use alternate covers
 
-        # If there is no URL return 100
-        if not primary_img_url:
+        # If there is no ImageHash or no URL and Hash, return 100 for a bad match
+        if primary_img_url is None or (primary_img_url.Hash == 0 and not primary_img_url.URL and not use_alt_urls):
             return Score(score=100, url="", remote_hash=0, local_hash=0, local_hash_name="0")
 
         self._user_canceled()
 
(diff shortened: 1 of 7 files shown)


mizaki:
Any issues with this? Seems like the easiest solution.

lordwelch:
The ImageHash is for making so we don't download the cover image, if a user needs it for manual matching we should instead give a link to the webpage for the comic.

Also adding url to `ImageHash` seems silly to me when we have `_alternate_images: list[str | ImageHash]`. Now if it's changed to `_alternate_images: list[ImageHash]` then it makes sense and overriding `__str__` is also silly we can just do `imagehash.URL`

mizaki:
ImageHash isn't to prevent the downloading of the image per se, it's to not do it if there is a hash for matching. For the issue window downloading the cover is fine (by Metron at least) as it's going to be far more limited. If the source doesn't wish to allow for the use of image download at all, the talker should leave it empty.

image
This is confusing when the alternative images will load (as there are no hashes for them and so normal URL strings are used).

It's the same problem with alternative images as they could be Hashes or URLs, so say, pushing the "main" cover URL to alternatives wouldn't work either.

Using the `__str__` means not having to do `if isinstance(obj, str)` so that `self.coverWidget.set_issue_details(self.issue_id, [str(issue._cover_image) or "", *alt_images])` from the `issueselectinwindow.py` for example can remain simple. Maybe it's preferred to break it out?
```python
cover_image_url = issue._cover_image if isinstance(issue._cover_image, str) else issue._cover_image.URL
```
Then `self.coverWidget.set_issue_details(self.issue_id, [cover_image_url or "", *alt_images])` from the `issueselectinwindow.py`

lordwelch:
> This is confusing when the alternative images will load (as there are no hashes for them and so normal URL strings are used).

That's on the talker for being inconsistent



> ```python
> cover_image_url = issue._cover_image if isinstance(issue._cover_image, str) else issue._cover_image.URL
> ```

Why?
If you're adding URL to `ImageHash`. Drop the string alternative
This way you never have to do any isinstance checks you just take the url attribute

mizaki:
> > This is confusing when the alternative images will load (as there are no hashes for them and so normal URL strings are used).
> 
> That's on the talker for being inconsistent
> 

The source may not provide hashes for alternatives (Metron being our only example currently) so it should be expected that alternatives may be a URL list and the cover is an `ImageHash`.

> > ```python
> > cover_image_url = issue._cover_image if isinstance(issue._cover_image, str) else issue._cover_image.URL
> > ```
> 
> Why? If you're adding URL to `ImageHash`. Drop the string alternative This way you never have to do any isinstance checks you just take the url attribute

`_cover_image` can be a string URL or an `ImageHash` depending on what the source provides. I don't see how the string URL can be dropped as it's required for `IssueIdentifier` (rather `ImageHash` is but then it presumes string) to know if it should check hashes or generate them?

Otherwise the talker would have to make all cover image URL into a hash and then I guess have the hash value be `0` or `None` and then `IssueIdentifier` would need to be changed to check for that but then that doesn't help with the problem of knowing when the image URL can be used for hashing or not.

Maybe it's something simple I'm missing here?

lordwelch:
Why would we download an image to hash it when we already have the hash?

lordwelch:
Download the image
```python
ImageHash(0, "", "https://image.jpg")
```
Use the hash
```python
ImageHash(23456, "ahash", "")
```
Use the hash
```python
ImageHash(23456, "ahash", "https://image.jpg")
```

I don't understand how it would be interpreted any other way

mizaki:
> Download the image
> 
> ```python
> ImageHash(0, "", "https://image.jpg")
> ```
> 
> Use the hash
> 
> ```python
> ImageHash(23456, "ahash", "")
> ```
> 
> Use the hash
> 
> ```python
> ImageHash(23456, "ahash", "https://image.jpg")
> ```
> 
> I don't understand how it would be interpreted any other way

All talkers would need to be changed. I presumed this was the reason for having the `ImageHash | str` so as to avoid that.

The edge case of being able to use the cover image URL in the GUI but not for matching would be ignored? It might help to prevent confusion I suppose as if you see the cover in the GUI but then it won't match...

I feel like other than the edge case there's something else but I can't think of it so maybe there isn't...

mizaki:
Shall I do something like:
```python
class GenericMetadata:
    def __setattr__(self, name: str, value: Any):
            if name == "_cover_image" and isinstance(value, str):
                self.__dict__["_cover_image"] = ImageHash(Hash=0, Kind="", URL=value)
            else:
                super().__setattr__(name, value)
```
And rename `ImageHash` to `ImageInfo`?

lordwelch:
No, plugins will need to update.
Either leave it `ImageHash` and put a comment in noting that if the hash/kind is empty then it will need to be calculated from the image at the url. If the url is empty then the hash/kind should not be empty. If both are empty it's invalid

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
         use_alt_urls: bool = False,
     ) -> Score:
         # local_hashes is a list of pre-calculated hashes.
-        # use_alt_urls - indicates to use alternate covers from CV
+        # use_alt_urls - indicates to use alternate covers
 
-        # If there is no URL return 100
-        if not primary_img_url:
+        # If there is no ImageHash or no URL and Hash, return 100 for a bad match
+        if primary_img_url is None or (primary_img_url.Hash == 0 and not primary_img_url.URL and not use_alt_urls):
```
lordwelch: Use Kind instead as 0 would still be a valid hash

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
-        else:
-            urls = [primary_img_url]
-            if use_alt_urls:
-                only_urls = [url for url in alt_urls if isinstance(url, str)]
-                urls.extend(only_urls)
-                self.log_msg(f"[{len(only_urls)} alt. covers]")
 
-            remote_hashes = self._get_remote_hashes(urls)
+        try:
+            if primary_img_url.Hash != 0 and primary_img_url.Kind:
```
lordwelch: Technically all values greater than -1 (up to and including 18446744073709551615) are valid hashes, using only the Kind is sufficient

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
+                remote_hashes.append((primary_img_url.URL or "", primary_img_url.Hash))
```
lordwelch:
```suggestion
                remote_hashes.append((primary_img_url.URL, primary_img_url.Hash))
```

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
+                self.log_msg(
+                    f"Using source talker hash for cover matching. Hash: {primary_img_url.Hash}, Kind: {primary_img_url.Kind}"
```
lordwelch:
```suggestion
                    f"Using provided hash for cover matching. Hash: {primary_img_url.Hash}, Kind: {primary_img_url.Kind}"
```

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
+                )
+            elif primary_img_url.URL:
+                remote_hashes = self._get_remote_hashes([primary_img_url.URL])
+                self.log_msg("Using URL download image for cover matching.")
```
lordwelch:
```suggestion
                self.log_msg("Downloading image for cover matching.")
```

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
+        except Exception as e:
+            logger.warning("Failed to process primary ImageHash: %s", e)
+
+        if use_alt_urls and alt_urls:
+            try:
```
lordwelch: What sort of exception was gotten here?

comictaggerlib/issueidentifier.py
```diff
@@ -306,35 +306,48 @@ def _get_remote_hashes(self, urls: list[str]) -> list[tuple[str, int]]:
...
+                only_urls = []
+                for alt_url in alt_urls:
+                    if alt_url.Hash != 0 and alt_url.Kind:
+                        remote_hashes.append((alt_url.URL or "", alt_url.Hash))
```
lordwelch:
```suggestion
                        remote_hashes.append((alt_url.URL, alt_url.Hash))
```

comictaggerlib/issueidentifier.py
```diff
@@ -525,13 +538,12 @@ def _match_covers(
             )
 
             try:
-                image_url = issue._cover_image if isinstance(issue._cover_image, str) else ""
                 # We only include urls in the IssueResult so we don't have to deal with it down the line
                 # TODO: display the hash to the user so they know a direct hash was used instead of downloading an image
-                alt_urls: list[str] = [url for url in issue._alternate_images if isinstance(url, str)]
+                alt_urls: list[str] = [img.URL or "" for img in issue._alternate_images]
```
lordwelch:
```suggestion
                alt_urls: list[str] = [img.URL for img in issue._alternate_images]
```

mizaki:
At the moment there is an image hash class for `quicktag` and `genericmetadata`. Should I consolidate them?

comictaggerlib/issueidentifier.py
mizaki: It was just a general catch but looking at it, it shouldn't be needed so I've removed the two `try`.

lordwelch:
> At the moment there is an image hash class for `quicktag` and `genericmetadata`. Should I consolidate them?

No that's going through a bunch of different changes right now

lordwelch:
Right now your tests are failing, but otherwise this looks good

PR #737, opened 2025-03-04
Initital CLI PluginUpdateManager

mizaki:
This is the basics of a plugin manager that can install and update plugins from a hand crafted list.

Anything terribly wrong before I continue?

Should the GUI be it's own new window or a new tab in settings?


comictaggerlib/__init__.py
```diff
@@ -1 +1,5 @@
 from __future__ import annotations
+
+import importlib.resources
+
+data_path = importlib.resources.files(__package__)
```
lordwelch, line 5: Not sure why this is here... any resources should be their own package

comictaggerlib/ctsettings/commandline.py
```diff
@@ -251,6 +251,30 @@ def register_commands(parser: settngs.Manager) -> None:
...
+        action="store_const",
+        const=Action.update_remote_plugins,
+        default=Action.gui,
+        help="Update all installed remote plugins.\n\n",
+        file=False,
+    )
+    parser.add_setting(
+        "--install-remote-plugin",
+        default="",
+        help="Install (or update) a remote plugin via its ID, eg. 'metronxml'. Use --list-remote-plugins for list.\n\n",
```
lordwelch, line 266: Right now IDs are not unique across different plugin types and Archivers don't even have IDs, they just get added to the list to be tried.

comictaggerlib/main.py
```diff
@@ -28,6 +28,7 @@
 from collections.abc import Collection
 from typing import cast
 
+import plugin_update_manager
```
lordwelch: This module isn't in this commit yet. Did you have something for this or was that next on your list?

lordwelch:
> Should the GUI be it's own new window or a new tab in settings?

I don't have a preference

comictaggerlib/main.py
mizaki: I may have forgot to `add` it...

I saw the comment about using YAML etc. for manifest but I just included what was supposed to be with the initial push.

And some mypy complaints to fix.

comictaggerlib/__init__.py
mizaki: I put it in to access the `plugin_list.json` but wasn't sure on the best way and best place to put the file. I copied the `publisher.json`.

comictaggerlib/ctsettings/commandline.py
mizaki: This would be the ID not from the plugin but the manifest.

mizaki:
This rework should be more like it.

I've used a test remote [manifest](https://gist.github.com/mizaki/52b60ac53cd3344ff1e1bfd44fc50ebd) YAML. I presume what I've done is what you meant? It'll be up to the plugin maintainer to generate the `manifest.yaml` file. It looks like it's possible with a github actions to create one on release.

comictaggerlib/ctsettings/commandline.py
lordwelch: We need to not use id then because it's used for a different purpose everywhere else.
You could rename id to name and only reference name, this is more of a convenience thing so I'm not interested in making things needlessly convoluted.

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                if not test_plugin:
+                    self._remove_plugin(new_plugin_file)
+
+            # Clean out old plugin files
+            self._remove_old_plugins()
+
+    def _check_for_all_updates(self) -> list[tuple[Plugin, str]]:
+        available_updates: list[tuple[Plugin, str]] = []
+        for k, item in self.plugin_files.items():
+            for r_plugin in self.remote_plugin_list:
```
lordwelch: I assume this would be separate in the final to allow custom plugins to point to their own manifest

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+    @staticmethod
+    def from_yaml(yaml_data: yaml.YAMLObject) -> PluginReleases:
+        downloads: list[Download] = [Download(**entry) for entry in yaml_data.get("downloads", [])]
+        return PluginReleases(latest=yaml_data["latest"], downloads=downloads)
+
+
+class PluginUpdateManager:
+    def __init__(self, config: ct_ns):
+        self.config = config
+        self.plugin_dir: Path = Path(self.config.Runtime_Options__config.user_plugin_dir)
```
lordwelch, line 86: It should use a separate directory for downloading new plugins, then you can remove the entire directory before updating that way leftovers are available if it fails

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+        try:
+            return parse(file)
+        except InvalidVersion:
+            logger.error(f"Invalid version number for : {file}")
+            return None
+
+    def _find_local_plugins(self) -> None:
+        local_plugins = plugin_finder.find_plugins(self.plugin_dir)
+
+        # Create a deduped dict with each sorted by Version number
```
lordwelch, line 226: All loading code should probably be handled in `plugin_finder.py` otherwise there can be inconsistencies

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                            break
+                else:
+                    # No "latest" in manifest, presume first download is latest
+                    success, new_plugin_file = self._download_plugin(latest_version.downloads[0].url)
+
+                if success:
+                    test_plugin = self._check_plugin(new_plugin_file)
+                    if test_plugin:
+                        print(  # noqa: T201
+                            f"Installed: {plugin_details.name} {latest_version.latest} to {self.plugin_dir}"
```
lordwelch: Use early returns instead of indenting further

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+            print(f"Failed to install plugin with ID '{plugin_id}', see log for details")  # noqa: T201
+
+    def _download_plugin_manifest(self, url: str) -> PluginReleases | None:
+        # TODO Remove test manifest URL
+        url = "https://gist.githubusercontent.com/mizaki/52b60ac53cd3344ff1e1bfd44fc50ebd/raw/f446fd7c149053965fe9e4035fe66e296bf3a02d/manifest.yaml"
+        latest = self._manifest_request(url)
+
+        try:
+            latest_yaml = yaml.safe_load(latest)
+            releases: PluginReleases = PluginReleases.from_yaml(latest_yaml)
```
lordwelch: You might as well put the `yaml.safe_load` call in `PluginReleases.from_yaml`

comictaggerlib/plugin_update_manager.py
```diff
@@ -0,0 +1,332 @@
...
+                    time.sleep(10)
+                    limit_counter += 1
+                    if limit_counter > 3:
+                        # Tried 3 times, inform user to check CV website.
+                        logger.error(f"{url} rate limit error. Exceeded 3 retires.")
+                else:
+                    break
+
+            except requests.exceptions.Timeout:
+                logger.debug(f"Connection to {url} timed out.")
```
lordwelch: Go with the simplest code possible and check whatever assumption is made, then let the user know what failed, they can retry themselves if they want to.

comictaggerlib/__init__.py
lordwelch: Give it it's own package like `publishers.json` is in it's own comicapi.data package

lordwelch (commented):
In general go with checking an assumption and failing early if it isn't what is expected

For example:
The manifest should always have a `latest` attribute (so we don't need to figure out what the latest version is) check that it has a latest attribute and fail loading the manifest if it doesn't

> I've used a test remote [manifest](https://gist.github.com/mizaki/52b60ac53cd3344ff1e1bfd44fc50ebd) YAML. I presume what I've done is what you meant? It'll be up to the plugin maintainer to generate the `manifest.yaml` file. It looks like it's possible with a github actions to create one on release.

Yes and then they can just link to it, they could even just generate it with only the current release and make it a release artifact and give the automatic link to the latest release EG https://github.com/comictagger/comictagger/releases/latest/download/ComicTagger-1.5.5-win-AMD64.exe.zip and replace `ComicTagger-1.5.5-win-AMD64.exe.zip` with `plugin-manifest.yaml` or whatever they feel like calling it

comictaggerlib/plugin_update_manager.py
mizaki: We can have a list as is here but also check the plugin itself for a `manifest` attribute and prefer that over the list?

```python
class KitsuTalker(ComicTalker):
    name: str = "Kitsu"
    id: str = "kitsu"
    manifest: str = "https://github.com/mizaki/kitsu_talker/blob/master/manifest.yaml"
```

comictaggerlib/ctsettings/commandline.py
mizaki: The two are now related since `plugin_finder.find_plugins()` is now used and the one ID in the manifest is checked against the plugin ID to make sure they match.

The manifest does have the `type` of plugin so the `id` and `type` could be checked.

Archivers will still require an `entry_name` which would be the ID? I guess the thing to do it take a builtin one and make it a plugin to check. 

Should external plugin archivers require an ID anyway?

comictaggerlib/plugin_update_manager.py
mizaki: A `download` dir in the `plugin` dir or use cache? Or a `tempfs` dir even?

As is, it is supposed that the new download file will have a different file name (version number etc.) and the new download is tested before the old one is removed. If the new one fails to load it is removed. But expecting a unique file name does seem primed for disaster.

Should the copied/moved file use the original file name or use a standard format so there is only ever the possibility of one file for one plugin (if using the manager and not copying things in manually)?

comictaggerlib/plugin_update_manager.py
mizaki: I'm not understanding this as it's using `plugin_finder`. 

After that it just extracts the `entry_name`, `path` and `version`.

comictaggerlib/ctsettings/commandline.py
mizaki: Testing an external archiver using the `entry_name` as a standin for ID looks to be reasonable?

comictaggerlib/__init__.py
mizaki: I've made a dir called `plugin_manifest` in `comictaggerlib`

comictaggerlib/plugin_update_manager.py
mizaki: It'll now check each plugin and if it has a class var called `manifest` it'll update the manifest URL if it's already present or add it.

comictaggerlib/plugin_update_manager.py
mizaki: Use a `download` dir in the `plugins` dir and move the old versions to `old` in the same.

mizaki:
I/GPT created this github action for on release: https://github.com/mizaki/kitsu_talker/blob/master/.github/workflows/generate-manifest-on-release.yaml

comictaggerlib/plugin_update_manager.py
```diff
@@ -63,120 +68,278 @@ class PluginReleases:
...
         self._find_local_plugins()
 
-    def _parse_version(self, file: str) -> Version | None:
+    def cli_list_available_updates(self) -> None:
+        """CLI method to list available updates of installed plugins"""
+        print("Checking for updates...")  # noqa: T201
+        available_updates: list[tuple[Plugin, str]] = self._check_for_all_updates()
+        if available_updates:
+            print("Available Update(s):")  # noqa: T201
+            [print(item[0].name) for item in available_updates]  # noqa: T201
```
lordwelch: don't put print in a list comprehension

comictaggerlib/plugin_update_manager.py
lordwelch: reuse the `local_plugins` loaded from `App.load_plugins`

lordwelch (commented):
Lets reduce the scope for the first iteration of this.
Lets just implement the listing of installable plugins and installing them, so no updates no pulling a list from other places just downloading a hardcoded plugin version from plugin_list.yaml

And assume that no one is managing plugins manually. When you think of a problem, (eg like "what if there's two copies in the plugin dir?") make the answer `raise Exception("Something failed")`
Also shutil and pathlib.Path cover a lot of platform weirdness eg shutil.rmtree, pathlib.Path.replace, pathlib.Path.mkdir(parents=True, exist_ok=True)
Also just make every path a pathlib.Path it makes it simpler to manage

mizaki:
I've cut it down to listing the default list and installing only.

I've had to leave the moving of older versions in because the plugin loaders do not make sure it's using the latest version.

lordwelch:
I'm not interested in working on this now.

Issue #740, opened 2025-03-20
AppImage not opening


N-Hertstein:
**Describe the bug**
Unable to open appimage

**To Reproduce**
1. Download appimage
2. Put it in the correct place with gearlever
3. Open it using desktop file
4. See it appearing in my taskbar and then vanishing

**Expected behavior**
Comictagger opens

**Attach logs**
No logs produced :(

**Screenshots**

**Desktop (please complete the following information):**
 - OS: Fedora 41 KDE Plasma
 - Version 1.6.0 (tried every beta with an appimage)
 - Downloaded the appimage straight from this github repo

**Additional context**
I just want to tag my comics :(

I might be missing a dependency, what does it use to display the window?

mizaki:
What happens if you run it from a terminal?

I just tried the latest with Fedora 41 KDE and it opened. (Yeah, I know, "works for me".)

You could also run the packaged normal linux https://github.com/comictagger/comictagger/releases/download/1.6.0-beta.4/ComicTagger-1.6.0b4-Linux.tar.gz

N-Hertstein:
Executing it from terminal worked somehow. Probably a desktop file issue? But executing it directy from dolphin with the left mouse button does not work? So who knows why terminal does work

It does say: "qt.qpa.wayland: EGL not available" when using terminal.
Seems like I'm missing some wayland libraries or don't have them properly configured. Will update what was missing :)
(But feel free to help me I'm not that familiar with linux libraries and stuff yet)

Eoin-ONeill-Yokai:
I'm having this issue as well when running in a terminal or from desktop. Here's the output log.

```
qt.glx: qglx_findConfig: Failed to finding matching FBConfig for QSurfaceFormat(version 2.0, options QFlags<QSurfaceFormat::FormatOption>(), depthBufferSize -1, redBufferSize 1, greenBufferSize 1, blueBufferSize 1, alphaBufferSize -1, stencilBufferSize -1, samples -1, swapBehavior QSurfaceFormat::SingleBuffer, swapInterval 1, colorSpace QSurfaceFormat::DefaultColorSpace, profile  QSurfaceFormat::NoProfile)
[...]

PR #746, opened 2025-03-29
Add error catching to locale.getlocale()

N-Hertstein:
Add error handling and fallbacks to en_US or C as language locale in case of misconfigured or minimal system.
This fixes issue #740


N-Hertstein:
My bad I'll use " in stead of '

PR #747, opened 2025-03-29
Add error catching to locale.getlocale(locale.LC_ALL, "")

N-Hertstein:
Add error handling and fallbacks to en_US or C as language locale in case of misconfigured or minimal system.
Fixes issue #740

### DIFF
--- a/comictaggerlib/main.py
+++ b/comictaggerlib/main.py
@@ -88,7 +88,7 @@ def configure_locale() -> None:
         if code != "":
             os.environ["LANG"] = f"{code}.utf-8"
 
-    #Get locale settings from OS, fall back to en_US or C in case of error for minimalist or misconfigured systems
+    # Get locale settings from OS, fall back to en_US or C in case of error for minimalist or misconfigured systems
     try:
         locale.setlocale(locale.LC_ALL, "")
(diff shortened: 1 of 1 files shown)


mizaki:
I don't know how possible it is but what if `C.UTF8` or `en_US.UTF-8` isn't available (`C` or `en_US` instead for example) etc.?

N-Hertstein:
> I don't know how possible it is but what if `C.UTF8` or `en_US.UTF-8` isn't available (`C` or `en_US` instead for example) etc.?

That would be better, fixed it.

mizaki:
What I mean is it's not known what will be on the system. So either the locales need to be interrogated and one picked or the best option might be to `pass` and log an exception.

lordwelch (changes requested):
Only use C as a fallback it will always exist and log that it's falling back to C `logger.error("Couldn't set the locale: unsupported locale setting; falling back to 'C' locale")`

N-Hertstein:
> Only use C as a fallback it will always exist and log that it's falling back to C `logger.error("Couldn't set the locale: unsupported locale setting; falling back to 'C' locale")`

Added as requested

PR #749, opened 2025-03-29
Sort plugins by version number

mizaki:
so the last plugin that will be loaded will be the latest version.

This seems to work with simple numbering but it may be save to parse to Version for the comparison.


lordwelch:
No, the user can manage that themselves. The only time that would reasonably happen is if it is a local plugin and they are already sorted before they are loaded for consistency

Issue #761, opened 2025-05-16
[Feature] GTIN field


Kijaru:
**Is your feature request related to a problem? Please describe.**
I'm using ComicTagger to update the metadata for my library, and I heavily rely on ISBN/EAN identifiers to check wether I already have an album or not. The main issue is that there's no way to set the identifiers directly from ComicTagger, one has to edit the ComicInfo manually to add it.

**Describe the solution you'd like**
As far as I understand, the ISBN or EAN belong in the GTIN field : https://anansi-project.github.io/fr/docs/comicinfo/documentation#gtin
Would it be possible to add this field to the interface and generated ComicInfo.xml? 

**Describe alternatives you've considered**
Adding the ISBNs manually into each ComicInfo

lordwelch:
Fixed in #762

PR #762, opened 2025-05-17
Add support for GTIN identifier

Kijaru:
This is my first public contribution ever, so please bear with me.

I've added support for the GTIN tag to represent the EAN/ISBN, as specified in https://anansi-project.github.io/fr/docs/comicinfo/documentation#gtin

Related issue: #761 

No tests were added, but one had to be changed to include the new identifier.

The text field is at the bottom of the Details page. 
The reading of the GTIN is implementation-dependant, so I figured a large text box would be fitting since one could use it with multiple identifiers at once (isbn:xxxxxx,ean:yyyyy,bookid:zzzzz).

### DIFF
--- a/comicapi/tags/comicrack.py
+++ b/comicapi/tags/comicrack.py
@@ -49,7 +49,6 @@ class ComicRack(Tag):
             "alternate_series",
             "alternate_number",
             "alternate_count",
-            "gtin",
             "story_arcs",
             "series_groups",
             "publisher",
@@ -233,7 +232,6 @@ class ComicRack(Tag):
         assign("AlternateCount", md.alternate_count)
         assign("StoryArc", md.story_arcs)
         assign("SeriesGroup", md.series_groups)
-        assign("GTIN", md.gtin)
 
         assign("Publisher", md.publisher)
         assign("Imprint", md.imprint)
@@ -321,7 +319,6 @@ class ComicRack(Tag):
         md.alternate_count = utils.xlate_int(get("AlternateCount"))
         md.story_arcs = utils.split(get("StoryArc"), ",")
         md.series_groups = utils.split(get("SeriesGroup"), ",")
-        md.gtin = utils.xlate(get("GTIN"))
 
         md.publisher = utils.xlate(get("Publisher"))
         md.imprint = utils.xlate(get("Imprint"))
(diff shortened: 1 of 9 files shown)


comicapi/tags/comicrack.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, version: str) -> None:
             "alternate_series",
             "alternate_number",
             "alternate_count",
+            "gtin",
```
lordwelch: This file is going to stay exclusively the original format that ComicRack (not ComicRack CE) read/writes.
This is the implementation for the Anansi project changes
https://github.com/comictagger/comicinfoxml

comictaggerlib/taggerwindow.py
```diff
@@ -843,6 +844,7 @@ def assign_text(field: QtWidgets.QLineEdit | QtWidgets.QTextEdit, value: Any) ->
         assign_text(self.leAltSeries, md.alternate_series)
         assign_text(self.leAltIssueNum, md.alternate_number)
         assign_text(self.leAltIssueCount, md.alternate_count)
+        assign_text(self.leGtin, md.gtin)
```
lordwelch: If you could add a validation that warns the user if an invalid gtin is used that would be useful. As officially only valid gtins should be used here. See https://github.com/anansi-project/comicinfo/issues/12#issuecomment-1439399165

You could probably accomplish using the signals from the QLineEdit `editingFinished, returnPressed, textChanged, textEdited`

lordwelch (changes requested):
> The reading of the GTIN is implementation-dependant, so I figured a large text box would be fitting since one could use it with multiple identifiers at once (isbn:xxxxxx,ean:yyyyy,bookid:zzzzz).

FYI, the gtin doesn't actually support multiple values and shouldn't have the type prepended.

I have one change needed with the comicrack.py
and a suggestion in taggerwindow.py

comicapi/tags/comicrack.py
Kijaru: Got it, I'll revert the changes made in this file.

comictaggerlib/taggerwindow.py
Kijaru: I've added a light red background when the GTIN isn't a valid one.
Validation is based on the control key, so it's only mathematical, because I doubt adding a check with a call to a GTIN database would be useful.

lordwelch (approved):

PR #796, opened 2025-10-05
Allow downloading from settings GUI

mizaki:
Both GCD and MangaBaka (Metron too maybe) allow for download a SQLite dump of the databases. This adds a button in the settings GUI that triggers a method in the talker to allow file download (to the CT cache dir).

Add download widget to GUI setting. 
Add download_file to ComicTalker.


lordwelch:
I don't really like this, I think leaving it manual or making it fully automatic is better. GCD also has anti-bot features to make you download the DB manually.

Also at a minimum this would need to have feature parity on the CLI. It's also currently abusing adding a setting that does nothing to essentially add a boolean property to the class that is then only checked outside of the class. Just add a boolean property or use `hasattr` to check that the function exists

Issue #805, opened 2025-11-13
Adding "tags" does not save when using CLI


dbarenholz:
**Describe the bug**
When using the CLI to add tags to a CBZ archive, the tags do not get saved to the `ComicInfo.xml` file. This is on the latest public release, codename Scanlan Shorthalt.

**To Reproduce**
1. Obtain a CBZ file. Let's assume it's called `file.cbz`
2. Add tags with the CLI, e.g. `comictagger -s -t cr -m 'tags=tag 1^,tag 2' file.cbz`
3. Open the `ComicInfo.xml` file that's in the archive
4. Notice an absence of `<Tags>`.

**Expected behavior**
If I run `comictagger -s -t cr -m 'tags=tag 1^,tag2' file.cbz`, I'd expect that said file's ComicInfo file contains those tags.

**Attach logs**
`~/.cache/ComicTagger/log` on Linux

**Desktop (please complete the following information):**
 - OS: `Archlinux, kernel 6.17.7`
 - Version: `ComicTagger 1.5.5`
 - Where did you install ComicTagger from? `yay -S comictagger`

dbarenholz:
Note that this issue exists both within `1.5.5` as well as with `1.6.0b6.dev0`.

lordwelch:
You want to use the CIX tags not the CR tags

PR #806, opened 2025-11-14
cli+fix: Allow saving and reading 'tags' for CR type ComicInfo

dbarenholz:
This commit adds 3 lines of code to the comicrack format so that CLI usage allows saving (and printing) "tags", which is supported by ComicInfo v2.1 (draft) according to the spec.

It does not add anything in the GUI.

### DIFF
--- a/comicapi/tags/comicrack.py
+++ b/comicapi/tags/comicrack.py
@@ -44,6 +44,7 @@ class ComicRack(Tag):
             "title",
             "volume",
             "genres",
+            "tags",
             "description",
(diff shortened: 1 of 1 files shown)


dbarenholz:
Fixes https://github.com/comictagger/comictagger/issues/805
Related to https://github.com/comictagger/comictagger/discussions/804

lordwelch:
You want to use the CIX tags they are included automatically in releases and hosted at https://github.com/comictagger/ComicInfoXml

PR #834, opened 2026-04-25
TPB Detection

lordwelch:
Attempts to detect TPB (Trade Paper Back) or other collected editions and match them appropriately

### DIFF
--- a/comictaggerlib/ctsettings/file.py
+++ b/comictaggerlib/ctsettings/file.py
@@ -63,6 +63,12 @@ def identifier(parser: settngs.Manager) -> None:
         type=int,
         help="ComicTagger will automatically add an additional cover that has any black borders cropped.\nIf the difference in height is less than %(default)s%% the cover will not be cropped.\ndefault: %(default)s\n\n",
     )
+    parser.add_setting(
+        "--tpb-detection",
+        default=False,
+        action=argparse.BooleanOptionalAction,
+        help="Enables TPB detection. May stop comics not marked as a TPB in ComicVine from being found.",
+    )
 
     parser.add_setting(
         "--sort-series-by-year",
--- a/comictaggerlib/ctsettings/settngs_namespace.py
+++ b/comictaggerlib/ctsettings/settngs_namespace.py
@@ -63,6 +63,7 @@ class SettngsNS(settngs.TypedNS):
     Issue_Identifier__series_match_identify_thresh: int
     Issue_Identifier__series_match_search_thresh: int
     Issue_Identifier__border_crop_percent: int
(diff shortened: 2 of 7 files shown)


lordwelch:
@professionaltart Try this pr it should work better than #436

professionaltart:
Thanks, just getting it now.  Let me know if this is a very bad wrong version of 'getting it' (for Win11):
- git clone repo
- checkout this pr
- install vs studio c++ build tools
- pip install .[all]
 - pip uninstall metron_talker
- %localappdata%\Python\pythoncore-3.14-64\Scripts\comictagger.exe

lordwelch:
I've got a better option for you. https://github.com/comictagger/comictagger/wiki/downloading-packages-from-pull-requests

professionaltart:
I missed that while looking at the wiki last week, sorry!

I know GH has changed a bit since you documented the process in 2022, but I can only get as far as clicking on [summary](https://github.com/comictagger/comictagger/actions/runs/24921890088?pr=834), where it says that, despite the PR action running, there are no artifacts.

lordwelch:
CI is running again this time it should have the artifact uploaded

professionaltart:
Thanks @lordwelch - the PR build appears to be working ok in general use.   My specific use-case was to not show a TPB as a possible match for a single issue, tho' - in the same way I'm using `--use-publisher-filter` to weed out Panini - which may be the inverse of the implied usage of `--tpb-detection`

lordwelch:
It should filter out TPBs based on if the current comic is a TPB. So that normal comics don't get matched with TPBs and TPBs don't get matched with normal comics

professionaltart:
I've started another tagging run, I'll check back later today and see if I can repro

professionaltart:
Thanks, could not repro on that limited run (~2000 files) using this:

.\comictagger.exe -sofa1it cix --use-publisher-filter --source comicvine  --cv-use-series-start-as-volume --tpb-detection --glob

lordwelch:
Are you able to upload the log? There should be some output related to the tpb detection

professionaltart:
I need to spend some time setting up fakevine, I think a couple of issues with the run earlier related to missing cache hits :(

professionaltart:
I've poked and prodded through another slice of the library, after deleting the local CV cache CT keeps, and sent you a link to a few more snippets of logs.  One suggestion was for a Volume One (a TPB by another name, so c'est la vie), and of the others, one of two were mis-tagged originally (which caused some confusion requiring manual intervention), and the others just offered up the TPB as an alternate option.

professionaltart:
Confirming things are working pretty well with that latest update!

PR #849, opened 2026-08-13
fix: fix metadata in columns

barrelltitor:
Wrong metadata was shown in the table
Name column had publisher
Title column had issue


lordwelch (approved):
I must have not been thinking very straight when I made 512b45b7f7a070b3521633ce84ea719d41fc58dd

